PR #11, opened 2020-12-15
Setup custom theming with Bootstrap

smcalilly:
## Overview
This PR configures Bootstrap with Sass so that we can have more customize theming in the application. This PR also has some small UI changes to the portfolio planner.

### Demo
ui-config-demo

### Notes
There is a tiny table refactor in here, which happened because I needed to add some custom selector components to the table rows.

## Testing Instructions
- Visit the review app: https://ccfp-asset-d-ui-styling-7jjaty.herokuapp.com/
- Make sure that you see the CCFP's dark green as a background color.
- Add/remove projects from the portfolio. The projects in the portfolio should be highlighted with a light green.


hancush (approved):
Looks good to me!

Issue #18, opened 2021-01-04
Build out the Priority Scoring Feature


smcalilly:
Here is the wireframe: https://projects.invisionapp.com/share/7FZ8N9VC4AT#/screens/439986471

This should be accessible from the project detail page.

smcalilly:
@fgregg I need some help on how to reconcile the wireframe and models.

The wireframe suggests that you can add a project to a project. See the "add project or place" button on this screen: https://projects.invisionapp.com/share/7FZ8N9VC4AT#/screens/439986470.

Based on user research, it seems like they will sometimes group multiple, similar projects into one project in order to get funding for all of the projects. For example, from my understanding, they'll group roofing repairs for several buildings into one "project", even though each building's repair project was originally its own separate "project". 

I'm guessing that this wireframe is trying to appease that requirement. [The "priority scoring" wireframe](https://projects.invisionapp.com/share/7FZ8N9VC4AT#/screens/439986471) appears to be doing the same thing, as evidenced by the table with multiple projects.

As the models currently stand, I don't understand how this might work from a code or UI or database perspective. There's some way we'll have to do this, but how?

fgregg:
Let's skip the "add project to project." 

They might really need a way of grouping project together, but i don't think that the "project" is the right concept for that.

smcalilly:
@fgregg I’m writing tests for PR #20 and found a bug. Whenever the code creates a new Project, it doesn’t create an associated ProjectCategory. Is that something that we’d want to include on the initial form, where a user creates a new project? If a ProjectCategory is something like "capital improvement" or "maintenance" etc, it might make sense to include that in the [create form for a project](https://projects.invisionapp.com/share/7FZ8N9VC4AT#/screens/438121592). 

Which do you think is the best choice? A project category can be either: 
1. created manually by a user when a new Project is created
2. created programmatically with default values when the Project is created

We have an issue for getting the ProjectCategory data from Garret here: #13, so I can create some dummy data for this in the meantime.

smcalilly:
Ignore my above comment. A project category would be created via the admin interface and not by an end user. For development fixtures, I'm using the project categories from [the csv of the CIP](https://github.com/datamade/ccfp-asset-dashboard/blob/master/raw/2021CIPTablesDRAFT10.28.2020updated_GW_clean.csv).

smcalilly:
Closed with #20.

PR #20, opened 2021-01-07
Added the Priority Scoring UI + Form

smcalilly:
## Overview
See PR title. This closes issue #18.

### Demo
priority-scoring

### Notes
This PR doesn't have full styling.

## Testing Instructions
- visit deployment https://ccfp-asset-d-priority-s-gc0kvb.herokuapp.com/projects/
- add a project
- make sure you can still update the project "overview" form
- update the priority score
- test the form validation. those fields can only be a number between 1 and 5.
- double check that the UI reflects your saved changes


fgregg:
@smcalilly how are you feeling about this tabbed form? seems like very low information density. 

maybe we just pull it into a single form and have a single "save" button for everything. we can then style it, with javascript, as a tabbed or single page.

smcalilly:
@fgregg yeah, each tab could be consolidated into one, especially because of how little space the priority scoring needs. It might also cut down on some code complexity with the views. Should I make that change?

smcalilly:
@fgregg I'm looking at how to achieve that, and I like this solution so much better: https://stackoverflow.com/questions/11101651/django-forms-having-multiple-models-in-meta-class#mb0.

fgregg:
yeah let's go for it.

smcalilly:
@fgregg I've refactored the form into one view, here is the commit: fe758d9. 

It uses [a django formset](https://docs.djangoproject.com/en/3.1/topics/forms/formsets/). I've broken the tests so I don't think a review app will deploy -- I'll work on fixing them in the morning.

fgregg:
ping me when the tests pass

smcalilly:
@fgregg Tests are passing and the form is much simpler now. There is a lot to improve with the UI, which I'll start working on and address in another PR if that's ok.

asset_dashboard/views.py
```diff
@@ -51,19 +53,42 @@ def post(self, request, *args, **kwargs):
 
         if form.is_valid():
             project = Project.objects.create(**form.cleaned_data)
+            project_score = ProjectScore.objects.get_or_create(project=project)
```
fgregg, line 56: why is this `get_or_create`? shouldn't it always be `create`?

fgregg, line 56: why is this get_or_create? shouldn't it always be create?

asset_dashboard/views.py
```diff
@@ -51,19 +53,42 @@ def post(self, request, *args, **kwargs):
...
+
+        score_form = context['score_form']
+
+        if form.is_valid() and score_form.is_valid():
+            form.save()
+            score_form.save()
+            messages.success(self.request, 'Project successfully saved!')
+            return super().form_valid(form)
+        else:
+            return super().form_invalid(self, form)
```
fgregg, line 94: why did this work? i thought this was what you were struggling with yesterday?

asset_dashboard/models.py
```diff
@@ -75,26 +75,28 @@ def __str__(self):
 
 
 class ProjectScore(models.Model):
+    min_value = 1
```
fgregg, line 78: i think i might think about taking this a step further and define 

`RangeValidators = [MinValueValidator(1), MaxValueValidator(5)]`

and then later.

```
core_mission_score = models.IntegerField(default=1, validators=RangeValidators)
```

fgregg (changes requested):
A few questions and a suggestion.

asset_dashboard/models.py
smcalilly: Neat. Will that allow me to get the class attribute `min_value` or `max_value` so that I can set those in the form rather than hardcoding them? See lines 30 and 31: https://github.com/datamade/ccfp-asset-dashboard/blob/44d696ec196881bfa63a91fafb4fea606d81d9f6/asset_dashboard/forms.py#L30. 

That's the reason I did this, because the model validators weren't making their way into the HTML outputted by the form class. Is there a better way?

asset_dashboard/forms.py
```diff
@@ -1,12 +1,32 @@
...
+            'ease_score',
+            'geographic_distance_score',
+            'social_equity_score'
+        ]
+
+    def __init__(self, *args, **kwargs):
+        super(ProjectScoreForm, self).__init__(*args, **kwargs)
+
+        for field_name, field in self.fields.items():
+            field.widget.attrs['min'] = ProjectScore.min_value
```
fgregg, line 30: i think you should be able to get this from the field object?

fgregg:
are you actually validating the project scores, or just depending on html5 checks?

asset_dashboard/forms.py
smcalilly: Not sure I understand. Are any of these answers near what you're talking about? https://stackoverflow.com/questions/3647805/get-models-fields-in-django

asset_dashboard/views.py
smcalilly: > are you actually validating the project scores, or just depending on html5 checks?

That's for the HTML validation.

asset_dashboard/forms.py
smcalilly: Any idea where to get these values from the field object? This is what is returned from a field [within the form class's](https://github.com/datamade/ccfp-asset-dashboard/blob/44d696ec196881bfa63a91fafb4fea606d81d9f6/asset_dashboard/forms.py#L29) `__init__`:
```
{'max_value': None, 'min_value': None, 'required': True, 'label': 'Social equity score', 'initial': 1, 'show_hidden_initial': False, 'help_text': '', 'disabled': False, 'label_suffix': None, 'localize': False, 'widget': <django.forms.widgets.NumberInput object at 0x7ff615486dc0>, 'error_messages': {'required': 'This field is required.', 'invalid': 'Enter a whole number.'}, 'validators': []}
```

fgregg: i'm surprised that `validators` is empty!

smcalilly: yeah it's a head scratcher

smcalilly: @fgregg Here is a commit with a different approach: 455308f. I'm not a big fan of that verbose `ProjectScore.__dict__[field_name].field.validators[0].limit_value` for a number of reasons:
1. what if the order of the array changes and `validators[0]` doesn't match `attrs['min']`? Is there a way to ensure that doesn't happen?
2. the naming isn't very descriptive
3. it's ugly

You mentioned a form field for the model. Is this what you were talking about? https://docs.djangoproject.com/en/dev/howto/custom-model-fields/#specifying-the-form-field-for-a-model-field

I think I see how that might work. If I created these form field for the ProjectScore form and model, shouldn't I also refactor the other form classes and widgets into a model form field? For the sake of consistency.

fgregg: i cleaned this up a bit with https://github.com/datamade/ccfp-asset-dashboard/pull/20/commits/ffae69de054b0d5b59f5c87347cce03af794400b

fgregg (approved):

fgregg:
@smcalilly you can go ahead and merge this

Issue #21, opened 2021-01-11
Nested forms for the Project model detail view


smcalilly:
A Project detail view needs different forms associated with it. [Based on this wireframe](https://projects.invisionapp.com/share/7FZ8N9VC4AT#/screens/439986470), we need to finish building out the form:
- [x] Project
- [x] ProjectScore (see #18)
<s>- [ ] ProjectFinances</s> (we don't need this)
<s> - [ ] Places (not sure the model?)</s> won't be in the model
- [x] Section 
- [x] ProjectCategory
- [x] various political districts
- [x] Zone
- [x] total budget for a project
- [x] Phase (currently a boolean value)

We also need to setup the admin section so that an admin can create some admin data. Some of these are mentioned in issue #13, but here's a more comprehensive list:
- [x] ScoreWeight

@fgregg am I missing anything here?

fgregg:
* project category
* political districts
* zone

all need to be in the form too.

smcalilly:
Before closing:
1. See note in PR #23 about the project categories.
2. I added the score weight as a fixture. The admins can update it however they want. If they add another score weight, the new score weight won't be added to the calculation. It will only use the one from the fixture. Need to communicate this for admin training.

smcalilly:
From Garret's notes about this form

Categories to add:

Estimated Bid Quarter. 
Values: Q1, Q2, Q3, Q4

Project Phase. 
Values: Phase 1, Phase 2, Phase 3 Engineering. Phase 3 Construction, Feasibility, Design, Construction

PR #23, opened 2021-01-20
Added a project category, phase completion, districts, and zones to the Project form

smcalilly:
## Overview
This PR adds the project category, phase completion, districts, and zones to the project form. It's based on the list in #21. There's a little bit of styling for the PR.

### Demo
project-districts

### Notes
I wanted to get this PR ok'd to make sure that I changed all of the district and project models correctly. I also need a check on how I'm doing the forms. 

The list from #21 has "Need" and "Score Weight" but I'm not sure how those fit into this form.

There are some tasks that I need to do to complete this PR and close out #21:
1. Test the formset and validation. The review app won't deploy with broken tests.
2. Add custom validation to the model fields (a regex to parse the input so that it matches "District 15" etc.)
3. Finish the javascript where a user can add an additional field to the districts.
4. Fix the bug where a user can clear out a district's name and save the form, which prompts an error. The error message renders on the screen but the deleted text is still missing. The fix for this is related to number 3 ^.

I had some issues making the project category fields with dynamic selections for the subcategory, so I made it show each category + subcategory within a drop-down selector. Those current options came from the raw CIP csv.

## Testing Instructions
The current code won't deploy a review app because the tests are failing on the new formsets.
- Pull the changes to your local computer
- Build the containers
- Load the fixtures: `docker-compose run --rm app python manage.py load_development_data` 
- Visit localhost:8000/projects/
- Add a project
- Update the project within the form
- Make sure that all of the parts of the form work, and not just the new fields.


fgregg:
failing with 

```
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 330, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 371, in execute
    output = self.handle(*args, **options)
  File "/app/asset_dashboard/management/commands/load_development_data.py", line 15, in handle
    SenateDistrict.objects.all().delete()
NameError: name 'SenateDistrict' is not defined
```

looks like you didn't import everything you need into the `load_development_data.py`

smcalilly:
> failing with
> 
> ```
>   File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 330, in run_from_argv
>     self.execute(*args, **cmd_options)
>   File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 371, in execute
>     output = self.handle(*args, **options)
>   File "/app/asset_dashboard/management/commands/load_development_data.py", line 15, in handle
>     SenateDistrict.objects.all().delete()
> NameError: name 'SenateDistrict' is not defined
> ```
> 
> looks like you didn't import everything you need into the `load_development_data.py`

oops, added back here: 3cd689f

smcalilly:
@fgregg Tests are passing and the review app has successfully deployed. You can test here: https://ccfp-asset-d-project-di-04qo1z.herokuapp.com/projects/

smcalilly:
@fgregg I've cleaned up the form code a little bit. I also added some javascript so a user can dynamically add senate districts. I think I'll be able to use a similar if we change the model relationships between the projects and districts.

fgregg:
Looking good!

* Zone
* House District
* Commission District

should all be foreign keys like Senate District

asset_dashboard/forms.py
```diff
@@ -1,17 +1,30 @@
...
 
-class ProjectForm(ModelForm):
+
+class StyledFormMixin(object):
+    """
+    Generic mixin to consistently style each form field's html <input> element.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(StyledFormMixin, self).__init__(*args, **kwargs)
```
fgregg, line 12: it's a bit better to do this as

```
```

that way we don't need to change anything if decide we want to change the class name.

asset_dashboard/models.py
```diff
@@ -166,28 +167,47 @@ class ProjectCategory(models.Model):
...
 
     def __str__(self):
-        return self.category
+        return f'{self.category} -- {self.subcategory}'
 
 
 class HouseDistrict(models.Model):
     name = models.TextField()
-    shape = models.PolygonField()
+    project = models.ForeignKey(Project,
```
fgregg, line 175: these should not  be foreign keys on house district, senate district, zone, and commissioner district.

they should be many-to-many relations on the project model.

fgregg (changes requested):
the form inputs for all the district types need to be drop down or othewised controlled vocabs, not free text forms

asset_dashboard/models.py
smcalilly: fixed with commit 9bc9149b0f23a29567f413cb4d3d3ac6b6ff69ec

asset_dashboard/forms.py
smcalilly: fixed with 9bc9149b0f23a29567f413cb4d3d3ac6b6ff69ec

asset_dashboard/models.py
```diff
@@ -166,28 +151,38 @@ class ProjectCategory(models.Model):
...
+
+    def __str__(self):
+        return self.name
 
 
 class CommissionerDistrict(models.Model):
-
     name = models.TextField()
-    shape = models.PolygonField()
+    # project = models.ForeignKey(Project,
```
fgregg, line 173: let's remove this dead code

fgregg (changes requested):
clean up some dead code, but otherwise looks great!

asset_dashboard/models.py
smcalilly: good catch. fixed here: 936caa2

fgregg (approved):

PR #27, opened 2021-01-27
Added a filterable data table + a JSON list view for the table

### LINKED ISSUE #26: Finish Project list view
The table needs:
- [x] show more information about a project
- [x] pagination
- [x] filtering
  - section
  - category
  - ~~districts?~~
  - ?
- [x] sorting

smcalilly:
## Overview
This PR improves the existing list view for the projects. 

Added:
- filtering by project category and project section
- searching by project name and project description
- sorting by column
- a json list view for all the projects
- table pagination

Closes #26.

### Demo
project-list-demo

### Notes
There is some basic styling and layout but this PR still doesn't fit the CCFP's style guide.

## Testing Instructions
- Visit https://ccfp-asset-d-project-li-qeexkx.herokuapp.com/projects/
- Add some projects. At least 11 so you can test the pagination.
- Test all of the filtering/searching.
- Test the column sorting.
- Make sure you can click on a row to visit the project's detail page.


asset_dashboard/models.py
```diff
@@ -151,7 +151,7 @@ class ProjectCategory(models.Model):
     subcategory = models.TextField(null=True)
 
     def __str__(self):
-        return f'{self.category} -- {self.subcategory}'
+        return f'{self.category} & {self.subcategory}'
```
smcalilly, line 155: @fgregg Need to talk about this with you because this is a workaround.

smcalilly:
@fgregg This is ready for review now.

asset_dashboard/models.py
```diff
@@ -149,9 +149,19 @@ class ProjectCategory(models.Model):
 
     category = models.TextField(null=False)
     subcategory = models.TextField(null=True)
+    slug = models.SlugField(max_length=250, null=True, blank=True)
```
fgregg, line 152: Couple things.

I led you astray when I called this a slug. Within django world, a slug is something that should be okay to appear in a url and so typically "consists of only letters, numbers, underscores or hyphens".

So sorry for that. Let's just call this "name" and have this be a "TextField"

Also, this should not be nullable or blankable?

asset_dashboard/views.py
```diff
@@ -39,9 +41,37 @@ def get_context_data(self, **kwargs):
...
 
+class ProjectListJson(BaseDatatableView):
+    model = Project
+    columns = ['name', 'description', 'section_owner', 'category', 'id']
+    order_columns = ['name', 'description', 'section_owner__name', 'category__category']
+    max_display_length = 500
+
+    def filter_queryset(self, qs):
+        search = self.request.GET.get('search[value]', None)
+        section = self.request.GET.get('columns[2][search][value]', None)
```
fgregg, line 60: hmm... is there another way to name these so we don't have to do "columns[2]"?

asset_dashboard/views.py
```diff
@@ -39,9 +41,37 @@ def get_context_data(self, **kwargs):
...
+        category = self.request.GET.get('columns[3][search][value]', None)
+
+        if search:
+            qs = qs.filter(Q(name__icontains=search) | Q(description__icontains=search))
+
+        if section:
+            qs = qs.filter(section_owner__name__icontains=section)
```
fgregg, line 67: this should be an exact match right? not an icontains?

fgregg (changes requested):
Overall looking really nice!

asset_dashboard/views.py
smcalilly: I'm pretty sure that's the way [the DataTables jquery plugin](https://datatables.net/examples/index) formats the query string for the request. 

If a user filters or searches the table, the plugin sends an ajax request to the server. The request includes query strings that are formatted like`columns[2]`.

Here's an example of what this query string looks like, from a test: https://github.com/datamade/ccfp-asset-dashboard/blob/c61812229a781f3b1c971fe8c7aa27801be6c54c/tests/test_views.py#L33

I'll look into the library and see if there's a way to make it more explicit.

asset_dashboard/views.py
smcalilly: Good point. Fixed here: 423a709]

asset_dashboard/models.py
smcalilly: I made it null/blank because I thought that an admin user would try and save without populating the slug. 

I worry about an admin user messing with that field, whenever they're adding categories within the admin section. Is there a way to prevent that from happening? Or, is there a way to add a message to that admin form that tells them what the field is for?

Here is the requested change: 106093f.

smcalilly: Ignore that last bit. I excluded the field from the admin: 54c2145.

fgregg (approved):

Issue #32, opened 2021-02-04
Total Score method


fgregg:
On the `ProjectScore` model define a method that takes the all the individual scores of a `ProjectScore` object, multiplies each one by the appropriate weight from `ScoreWeights` and returns a total score

something like

```python
def total_score(self):
     score_weights = ScoreWeights.objects.get()
     total_score = 0
     for field in self:
          total_score += field * score_weight['appropriate weight']

     return total_score
```

Handle the case where we are missing an individual score by setting the total score to 0

fgregg:
@smcalilly let me know if you need any more detail on this

smcalilly:
@fgregg Two questions:
1. Can you think of any DataMade code that has methods on the model like this? An example would be helpful -- I'm having some issues working with the model instance.
2. Will the ScoreWeights be something that is added in the admin section? And will this only have one row in the database? My work-progress-code has this pk hardcoded:
```
score_weights = ScoreWeights.objects.get(pk=1)
```

fgregg:
>     2. Will the ScoreWeights be something that is added in the admin section? And will this only have one row in the database? My work-progress-code has this pk hardcoded:
> 
> 
> ```
> score_weights = ScoreWeights.objects.get(pk=1)
> ```

correct!

smcalilly:
@fgregg you said: "Handle the case where we are missing an individual score by setting the total score to 0". 

Each field in the ProjectScore model has `default=1`, so the score won't be missing. Should I remove that default or change it to 0? 

Here is my code:

```
class ProjectScore(models.Model):

    project = models.OneToOneField(Project, on_delete=models.CASCADE)
    core_mission_score = models.IntegerField(default=1,
                                             validators=[MinValueValidator(1),
                                                         MaxValueValidator(5)])
    operations_impact_score = models.IntegerField(default=1,
                                                  validators=[MinValueValidator(1),
                                                              MaxValueValidator(5)])
    sustainability_score = models.IntegerField(default=1,
                                               validators=[MinValueValidator(1),
                                                           MaxValueValidator(5)])
    ease_score = models.IntegerField(default=1,
                                     validators=[MinValueValidator(1),
                                                 MaxValueValidator(5)])
    geographic_distance_score = models.IntegerField(default=1,
                                                    validators=[MinValueValidator(1),
                                                                MaxValueValidator(5)])
    social_equity_score = models.IntegerField(default=1,
                                              validators=[MinValueValidator(1),
[...]

PR #33, opened 2021-02-04
added logic to the total_score method + tests for the method

smcalilly:
## Overview
This PR has 
- a model method for totaling the score for a project. The calculations are based on a project's ProjectScore and the ScoreWeights.
- tests for the method
- a UI change to show the total score on the project detail page
- a DB migration that missed out on PR #24...

It's for issue #32.

### Demo
Screen Shot 2021-02-04 at 3 05 47 PM

### Notes
See my note in the issue: https://github.com/datamade/ccfp-asset-dashboard/issues/32#issuecomment-773553971. 

## Testing Instructions
- Visit the review app and see the score on a project detail page.
- Change the score in the UI. When you save, it should be updated.

- Pull down the code locally and run the tests: `docker-compose -f docker-compose.yml -f tests/docker-compose.yml run --rm app`

- You can also test in the console. You'll first need to load the fixtures so the score weight exists: `docker-compose run --rm app python manage.py loaddata asset_dashboard/fixtures/data.json`
- Here is some code I was copy/pasting into the app's python shell while writing this method:
```python
from asset_dashboard.models import ScoreWeights, Project, ProjectScore
project = Project(name="new trailhead", description="we need a new trailhead")
new_score = ProjectScore(project=project, core_mission_score=3, operations_impact_score=4, sustainability_score=2, ease_score=5, geographic_distance_score=2, social_equity_score=4)
new_score.total_score
# prints 14.3
```
- Check the math


asset_dashboard/models.py
```diff
@@ -64,33 +64,42 @@ def __str__(self):
...
+    social_equity_score = models.IntegerField(null=True, blank=True,
                                               validators=[MinValueValidator(1),
                                                           MaxValueValidator(5)])
 
     @property
     def total_score(self):
-        '''return the total, weighted score'''
-        ...
+        # get the ProjectScore fields, excluding the id and project fields
+        score_fields = [f for f in self._meta.get_fields() if f.name != 'id' if f.name != 'project']
```
fgregg, line 89: this is a good solution!

i think we have an opportunity to do something a bit safer and much clearer.

let's define a custom fieldtype that looks like

```python
class ScoreField(models.IntegerField):

          if 'null' not in kwargs:
              kwargs['null'] = True
          if 'blank' not in kwargs:
               kwargs['blank'] = True
          if 'validators' not in kwargs:
               validators = [MinValueValidator(1), MaxValueValidator(5)]
          super().__init(self, *args, **kwargs)

class ProjectScore(models.Model):

     ...
     core_mission_score = ScoreField()
     ...

     @property
     def total_score(self):
          score_fields = [f for f in self._meta.get_fields() if type(f) == ScoreField]
          ...
```

Does that make sense?

asset_dashboard/models.py
```diff
@@ -64,33 +64,42 @@ def __str__(self):
...
+        score_weights = ScoreWeights.objects.all()[0]
```
fgregg, line 90: does it work to do `ScoreWeights.objects.get()`

if so, that would be preferred because it will be raises a useful error if there is more than one score weight

asset_dashboard/models.py
```diff
@@ -64,33 +64,42 @@ def __str__(self):
...
+        total_score = 0
+
+        for field in score_fields:
+            score_field_value = field.value_from_object(self) if field.value_from_object(self) else 0
+            weight_field_value = field.value_from_object(score_weights)
```
fgregg, line 95: wow, i didn't think that would work.

asset_dashboard/models.py
```diff
@@ -64,33 +64,42 @@ def __str__(self):
...
     def total_score(self):
-        '''return the total, weighted score'''
-        ...
+        # get the ProjectScore fields, excluding the id and project fields
+        score_fields = [f for f in self._meta.get_fields() if f.name != 'id' if f.name != 'project']
+        score_weights = ScoreWeights.objects.all()[0]
+        total_score = 0
+
+        for field in score_fields:
+            score_field_value = field.value_from_object(self) if field.value_from_object(self) else 0
```
fgregg, line 94: if a single field element is undefined, then the *total score* should be 0

fgregg (changes requested):
very nice work! here are some suggestions and clarifications

asset_dashboard/models.py
smcalilly: real cool, i'm about to try it. thanks.

asset_dashboard/models.py
smcalilly: that's what i originally had, then my tests failed because that pk didn't exist. is there a way to get the primary key, dynamically?

asset_dashboard/models.py
smcalilly: ah, yeah. that makes sense.

asset_dashboard/models.py
fgregg: not off the top of my head. let's do it this way

```python
# there should be one, and only one row in score weights. 
# this tuple unpacking will throw an error if that's not so
score_weights, = ScoreWights.objects.all()
````

asset_dashboard/models.py
smcalilly: Good DRY code at this new commit: 8822d65

asset_dashboard/models.py
smcalilly: cool, today i learned about the tuple unpack error. saving this snippet. commit here: 970e47e

asset_dashboard/models.py
fgregg: nice!

asset_dashboard/models.py
smcalilly: fixed here: d11687a

asset_dashboard/models.py
smcalilly: improved the test here: 5d3d220

fgregg (approved):

Issue #29, opened 2021-02-01
Improve the CIP planner


smcalilly:
The CIP planner is configured to use dummy data. We need to make changes so that we're using real data and the planner can be more useful.

- [x] Adapt the current code to use the application's real models.
- [x] Need to add a "total budget" input for the Project form. The calculator is dependent on this.
 
- [x] Create a way to save a potential plan. Three options:
  1. Create a form where a user can click "save" and it saves the potential plan. This would need at least a day or two of work. The work would require:
      - Django form + view changes
      - a popup modal so a user can enter the name of the plan + submit the form
      - Tests for the new view
      - List view for all the plans
      - Adapt the existing planning calculator/table for a detail view.
      
  2. Add some logic in the UI where a user can click "export as CSV". This can all happen in the browser and wouldn't have any data persistence on the server. This is the quickest and easiest way -- would take 3 - 6 hours.
  3. Combine i & ii to make two useful features.

Updated, based on notes from Garret:
- [x] Be able to sort by Total Score or Total Budget in the All Projects list.
- [x] Add Phase as another column
- [x] Click on project to take you back to project detail page.
- [x] Export Project Portfolio to CSV or Excel (one or the other)

PR #38, opened 2021-02-08
CIP Planner improvements for #29

smcalilly:
## Overview
- Adds a CSV export for the CIP planner
- Changes the view that sends the data to the CIP planner
- Adds Phase as another column to the planner
- Sorting for the total score and total budget on the table
- Adds a button for each project in the planner, so a user can see the project detail page
- Refactors some styling and improved the look -- made the tables take up more of the blank space
- Reorganizes some of the file/folder locations for CSS/JS libraries

Closes issue #29.

### Demo
planner-improvements

## Testing Instructions
- visit the review app deployment
- add a project
- update the project with a budget, a phase, scores, districts, and zones
- go to the cip planner. make sure all that data shows up in the table and the csv
- test the csv export
- test the sorting in the table


asset_dashboard/static/js/components/ProjectsTable.js
```diff
@@ -11,8 +11,12 @@ const ProjectsTable = ({ allProjects, columns, onAddToPortfolio }) => {
   }
 
   return (
-    <div className="mt-4">
-      <h3>All Projects</h3>
+    <>
```
fgregg, line 14: what does this empty tag do?

asset_dashboard/views.py
```diff
@@ -12,13 +13,34 @@
...
     def get_context_data(self, *args, **kwargs):
         context = super().get_context_data(*args, **kwargs)
-        projects = serializers.serialize('json', Project.objects.all())
-        context['props'] = {'projects': projects}
+
+        projects = Project.objects.all()
+        projects = projects.select_related('section_owner', 'category', 'projectfinances', 'projectscore')
+
+        projects_list = []
+        for prj in list(projects):
```
fgregg, line 26: i don't think you need to cast to a list here.

fgregg (changes requested):
looks good. one style issue and otherwise good to go

asset_dashboard/static/js/components/ProjectsTable.js
smcalilly: It's a [React Fragment](https://reactjs.org/docs/fragments.html#short-syntax). Basically, JSX requires sibling HTML elements to be wrapped in a parent tag. 

Normal HTML might look like:
```html
 // parent div
<div>
  // sibling #1
  <h1>Hello, I'm a Dog</h1> 

   // sibling #2
  <p>Will you please take me outside?</p> 
<div>
```

The above would also work in React. But if you wanted to remove the parent `div`, then this would cause an error in React:
```jsx
  // no parent element wrapping the two sibling elements
  <h1>Hello, I'm a Dog</h1> // sibling #1
  <p>Will you please take me outside?</p> // sibling #2
```

So instead of having div soup, I thought it would be better HTML to use a React Fragment:
```jsx
<> // parent element
  <h1>Hello, I'm a Dog</h1> // sibling #1
  <p>Will you please take me outside?</p> // sibling #2
</>
```

fgregg: got it! thanks.

asset_dashboard/views.py
smcalilly: Fixed here: 6add908

fgregg (approved):

Issue #41, opened 2021-02-17
Add a page for filtering projects by Senate, House, and Commissioner District


smcalilly:
In Garret's email, he said we need to add a page for filtering projects by Senate, House, and Commissioner District. The data needs to be exportable as a CSV.

PR #42, opened 2021-02-17
Created a basic view, template, and url for #41

smcalilly:
## Overview
This is the first step for #41. The implementation is basic and the feature incomplete.

The PR includes:
- a new view for listing projects that are filterable by district
- url for the view
- a link in the menu for the new url
- template for the view

### Demo
district-list-view

### Notes
I was thinking that I would do this filtering like I've done with the project list view. I used the [datatables jquery plugin](https://datatables.net/) for the UI and [django-datatables-view](https://pypi.org/project/django-datatables-view/) for the Django view.

The districts are hardcoded in the table, so they aren't using real data. There's a note in the code about that.

I'm not sure what to call this, especially since a project list view already exists (with filtering). I landed on "projects by district". We could stuff this functionality into the existing project list table but it might be too much in one table...thoughts on this?

## Testing Instructions
- Visit the deployment. You'll need to add a project so one shows up in the new table.
- In the dropdown menu, click on the "Projects by Districts"
- See the table with the projects.

### DIFF
--- a/asset_dashboard/templates/asset_dashboard/projects_by_district_list.html
+++ b/asset_dashboard/templates/asset_dashboard/projects_by_district_list.html
@@ -32,10 +32,12 @@
                   <td>{{ project.category }}</td>
 
                   <!-- TODO: show the list of district names. 
-                    i need advice on how to do this. each district returns a queryset.
-                    somehow, a project's list of districts need to be combined for display
-                    in the table. it could look like: Senate 17, Senate 18 or 17, 18.
-                    should i parse out each district in the template or create a model method to do this? -->
+                    i need advice on how to do this. each district returns a queryset, so
+                    a project's list of districts need to be combined for display
+                    in the table. it could look like either: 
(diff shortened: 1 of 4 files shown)


asset_dashboard/templates/asset_dashboard/projects_by_district_list.html
```diff
@@ -0,0 +1,51 @@
...
+                  <td>{{ project.description }}</td>
+                  <td>{{ project.category }}</td>
+
+                  <!-- TODO: show the list of district names. 
+                    i need advice on how to do this. each district returns a queryset, so
+                    a project's list of districts need to be combined for display
+                    in the table. it could look like either: 
+                        Senate 17, Senate 18
+                        17, 18
+                    to do this, should i parse out each district in the template or create a model method? -->
```
fgregg, line 40: i say parse it in the template. '17,18' looks better to me.

fgregg (approved):

PR #70, opened 2021-12-02
Configure second GIS database and router, add example model and skeleton models for remainder of quercus schema

### LINKED ISSUE #60: How do we want to connect to the FPDCC’s GIS Database
There seem like two reasonable options

1. Set up [more than one database connection in the Django settings](https://docs.djangoproject.com/en/3.2/topics/db/multi-db/).
2. Set up [foreign data wrapper](https://www.postgresql.org/docs/9.5/postgres-fdw.html) in our primary database to the GIS database

The first path has the benefit of being a bit more standard, so we can probably find better documentation and examples to follow.

The second path is going to be a *lot* more efficient for doing spatial queries, because we can do that all within a single SQL query instead of serializing data from the GIS data moving it through Python and putting it into getting it into the main database for querying.

We could also set up a synch between the GIS database and the main database and copy the pieces we need. I would really like to avoid that because that's just another thing to get out wack.

hancush:
## Overview

See title.

Closes #60

### Notes

AFAICT, the currents tests are OK, but heads up that pytest-django [only has experimental support for multiple databases](https://pytest-django.readthedocs.io/en/latest/database.html#tests-requiring-multiple-databases).

## Testing Instructions

* Run the app and load the fixture data and GIS data dump, as documented in the README.
* Open up a Django shell.
* Query for a managed model (e.g., `Section`) and confirm you get hits: `Section.objects.all()`
* Query for a `NaturePreserves` and confirm you get hits: `NaturePreserves.objects.all()`


asset_dashboard/models.py
```diff
@@ -248,3 +248,205 @@ class DummyProject(models.Model):
     def __str__(self):
         """"String for representing the model object"""
         return self.name
+
+
+class NaturePreserves(models.Model):
+
+    class Meta:
+        managed = False
+        db_table = '"quercus"."nature_preserves"'
```
hancush, line 257: Needed to quote the schema and table names for this to work. Not sure why, also lost StackOverflow reference, sorry!

asset_dashboard/models.py
```diff
@@ -248,3 +248,205 @@ class DummyProject(models.Model):
...
+
+    id = models.AutoField(primary_key=True, db_column='nature_preserves_id')
```
hancush, line 259: When a primary key isn't explicitly specified, Django tries to use `id`. There isn't an `id` column in many of these tables, so you have to specify it in `db_column`.

asset_dashboard/routers.py
```diff
@@ -0,0 +1,49 @@
+class GISRouter:
```
hancush, line 1: This tells Django which DB to route ORM queries to. See: https://docs.djangoproject.com/en/3.2/topics/db/multi-db/#automatic-database-routing

asset_dashboard/models.py
```diff
@@ -248,3 +248,192 @@ class DummyProject(models.Model):
...
+    site_name = models.CharField(max_length=254)
+
+    def __str__(self):
+        return self.site_name
+
+
+class Buildings(GISModel):
```
hancush, line 277: @smcalilly roughed in models for the remaining tables. The `db_table` attribute needs to be updated to quote the schema and table names, and the columns we need to use need to be defined as attributes.

tests/test_router.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+
+from django.db.utils import ProgrammingError
+
+from asset_dashboard.models import Section, NaturePreserves
+
+
+@pytest.mark.django_db(databases=['default', 'fp_postgis'])
+def test_gis_query(nature_preserves):
+    assert NaturePreserves.objects.db == 'fp_postgis'
```
hancush, line 10: If `using` is not passed, this attribute is set by the `db_for_read()` method of the router. See: https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/db/models/manager.py#L135

fgregg (approved):
looks really good, @hancush !

asset_dashboard/models.py
```diff
@@ -248,3 +248,192 @@ class DummyProject(models.Model):
...
+
+    class Meta(GISModel.Meta):
```
smcalilly, line 279: oh cool, i was wondering how to inherit so i didn't have to add `managed = False` to every model.

hancush: Yeah! The class-on-class Meta inheritance is a little wonky, but [it's buried down there in the docs](https://docs.djangoproject.com/en/3.2/topics/db/models/#meta-inheritance). (I had to Google it 😅)

PR #82, opened 2021-12-06
Add Phase model

hancush:
## Overview

This PR adds a `Phase` model. A phase is a sub-project unit of work.

This PR intentionally omits material changes to the Project creation / editing interfaces. Will do in #80. It does update some view code so the related Phase and PhaseFinances objects are created when a project is added, but this implementation should not be considered final. I've left TODOs in places I've commented or otherwise made stopgap changes to the existing code.

Related to but doesn't quite close (?) #77.

## Testing Instructions

* This PR provides tests for the new data migration, both forwards and backwards. Review those tests to confirm I've implemented the data migration as we like, then confirm CI passes.

PR #83, opened 2021-12-13
Add models for the GIS database

### LINKED ISSUE #61: Create managed django models for the GIS data
https://docs.djangoproject.com/en/3.2/ref/models/options/#managed

smcalilly:
## Overview
This PR adds the models for the `quercus` schema. For #61.

Includes:
- Unmanaged models with the relationships.
  - I followed what I could observe in my local database. See the schema from postgres for quick reference (see `postgis_schema.md`).
  - I added docstrings from [their documentation](https://github.com/fpdcc/triggers_and_updates/blob/master/documentation.md).
- Tests for the models and relationships.
  - I added a new docker command to the README so we can test the model relationships when the tests run in a local development environment.
- Pinned requirements after issues with Django 4.0. I copy/pasted the requirements in PR #82. 

### Notes
I wasn't sure how to use the local `fp-postgis` database without adding a boolean flag to the docker-compose command. I don't know if this is anti-pattern or what -- couldn't figure out how to combine the existing code in `conftest.py` with the local `fp-postgis` docker container. I thought creating test fixtures would be unnecessary work since we already have the data we need in that container's database. These tests won't run in a CI/CD action. They're meant to run only in a developer's local environment, with the `fp-postgis` database. Let me know if there's a better way to do this (cc @hancush)

Any ideas on how to make the tests more DRY?

Should we move these models to their own file? If so, what would be a good name?

## Testing Instructions
- Pull down this branch `git pull origin unmanaged-models && git checkout unmanaged-models`
- Run the tests: `docker-compose -f docker-compose.yml -f tests/docker-compose.yml run -e TEST_GIS=True --rm app`.
  - Make sure that you have the `fp-postgis` database loaded (see README instructions if not).
- Make sure the models are implemented the right way based on the schema located at `postgis_schema.md`.


fgregg (approved):

fgregg:
looks good to me maybe @hancush has a better idea about how to handle the test settings. i don't right now.

requirements.txt
```diff
@@ -1,21 +1,17 @@
...
-Django
-psycopg2-binary
-gunicorn
-dj-database-url
-whitenoise
-django-compressor
-sentry-sdk
-django-money
-django-datatables-view
+Django==3.1.7
```
hancush, line 1: Can you confirm these requirements match those pinned in #82, which I generated from the production deployment, except where noted? https://github.com/fpdcc/ccfp-asset-dashboard/pull/82/files#r764218837

tests/test_gis_models.py
```diff
@@ -0,0 +1,273 @@
...
+
+import pytest
+
+from asset_dashboard import models
+
+
+def get_all(model):
+    """Helper function since every test gets a queryset.
+    Also that the qs isn't empty."""
+    if bool(os.environ.get('TEST_GIS')):
```
hancush, line 13: This would be more appropriate to implement in a `skipif` mark: https://docs.pytest.org/en/6.2.x/reference.html#pytest-mark-skipif

requirements.txt
smcalilly: Yeah I copy/pasted from your file. Just double checked and it looks right.

tests/test_gis_models.py
smcalilly: Cool, I added that here: f8c2408

I combined the pytest decorators into a single decorator since they're used for every test and both pertain to the gis database. Much cleaner than my original implementation. Let me know if you see any room for improvement.

smcalilly: Thanks to this SO: https://stackoverflow.com/a/44635908

requirements.txt
hancush: Thanks!

hancush (commented):
I don't love that this solution means we aren't testing the GI models automatically, leaving room for us to introduce bugs we might not notice until we deploy the app. @smcalilly Can you say more about why you decided to test against the loaded dump, instead of following the usual pattern of creating test fixtures via the ORM, as in #70?

smcalilly:
@hancush These tests were meant more as a guide for code review and development work so we can understand the GIS data at a glance, as well as make sure the relationships work when accessed with the ORM. I asked in the slack channel if it was okay not to add test fixtures and Forest said that it was okay, but maybe my question wasn't clear enough. With how involved these models and relationships are, it seems like a lot of unnecessary work to recreate all these relationships and create the geo data in a pytest fixture (for now, see second paragraph). It especially seems like extra work that could be avoided since the data with all the relationships already exists in a local database, and we're not responsible for creating that data/we're using it as a read-only database. Also, writing out pytest fixtures for all the models might not be necessary at this point in time since we don't know what all models, fields, and relationships we'll need to use at this point in time.

I agree that it could open us up to bugs, but I don't think that would be an issue until we start making relationships with the existing models (the managed models created by our team, like `Project`, `Task`, etc). Though that could be wrong if you have more insight? For example, if we start relating a `Project` or `Task` model with one of the unmanaged GIS models, then I think it would be a good idea to always test those relationships during deployment, and would be worth the time to write out fixtures for the pieces that interact.

hancush:
@smcalilly I buy this:

> Also, writing out pytest fixtures for all the models might not be necessary at this point in time since we don't know what all models, fields, and relationships we'll need to use yet... if we start relating a Project or Task model with one of the unmanaged GIS models, then I think it would be a good idea to always test those relationships during deployment, and would be worth the time to write out fixtures for the pieces that interact.

Thanks for expounding. Can you remove the commented code from your test method? Then this has my approval.

tests/test_gis_models.py
```diff
@@ -0,0 +1,295 @@
...
+    return qs
+    # if bool(os.environ.get('TEST_GIS')):
+    #     qs = model.objects.all()
+    #     assert qs.count() > 0
+    #     return qs
+    # else:
+    #     # skip the test if no TEST_GIS env variable
+    #     # kinda hacky / workaround !
+    #     # this function doesn't follow the SRP...
+    #     pytest.skip()
```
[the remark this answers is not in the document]
smcalilly: whoops! got distracted by implementing the decorator. changed here: 2d7b1d3

Issue #80, opened 2021-12-06
Design interface for creating bundled projects


fgregg:
Needs to be near the interface for creating individual projects.

PR #84, opened 2021-12-17
add new phase model to the UI form

smcalilly:
## Overview
This PR puts the project edit form back into a working state after the changes from PR #82. It's related to #80 but not quite the work for that. There is no other issue for this.

Changes include:
- new `PhaseForm` class for the `Phase `model
- captured the form in two new views:
  -  `ProjectCreateView`
  -  `ProjectUpdateView`
- new template for the phase form
  - since the phase is a child to a project, the template is accessible through a project
- refactored the project template to the partial `project_base.html` so I can reuse those UI elements for this phase form and later, for the asset UI
- renamed the `project_form.html` partial to the more descriptive `edit_project_form.html`
- updated the test to capture the correct phase data

### Demo
phase-redone

## Testing Instructions
- Visit review app: https://ccfp-asset-d-phase-form-kaks9x.herokuapp.com
- Add a phase to a project.
- Make changes to a project's phase.


asset_dashboard/views.py
```diff
@@ -64,7 +64,7 @@ class ProjectListView(LoginRequiredMixin, ListView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        # send the form to a modal in this view
+        # send a new Project form to the modal in this view's template
         form = ProjectForm()
```
smcalilly, line 68: was confused by this at first but then i remembered how this works in the ui. in the project list view (the home page), a new project is added in a modal form. the new project has some very basic details then the user is redirected to the project edit page, where they can fill out more details.

asset_dashboard/views.py
```diff
@@ -164,7 +164,8 @@ def form_valid(self, form):
         forms = {
             'project': form,
             'score': context['score_form'],
-            'finances': context['finances_form']
+            'finances': context['finances_form'],
+            'phase': context['phase_form']
         }
```
smcalilly, line 169: when we did this last year, i thought this was kind of weird. but now, after working with it, i think this pattern gives us more flexibility for working with these different forms within the ui.

smcalilly:
@fgregg this is ready for re-review (there's no button to click for that because it's still waiting from the last iteration)

fgregg (approved):

PR #90, opened 2021-12-20
add phase model to the CipPlannerView

smcalilly:
## Overview
Fixes the CIP Planner, which was broken after we added the new `Phase` model. There's no issue for this.

## Demo
cip-planner-fix

## Notes
Is there a better way to deal with the `Project`/`Phase` relationships using the ORM? `select_related` and `prefetch_related` wouldn't work, so I ended up querying it like this code implementation. Seems like that would be inefficient with a ton of unnecessary database queries?

## Testing Instructions
- visit review app: tk
- create project so there's some data to test with
- (it won't have the phase stuff since that form isn't implemented?)
- go to the cip planner and make sure it works

### DIFF
--- a/asset_dashboard/views.py
+++ b/asset_dashboard/views.py
@@ -24,25 +24,28 @@ class CipPlannerView(LoginRequiredMixin, TemplateView):
         context = super().get_context_data(*args, **kwargs)
 
         projects = Project.objects.all()
-        projects = projects.select_related('section_owner', 'category', 'phasefinances', 'projectscore')
+        projects = projects.select_related('section_owner', 'category', 'projectscore')
 
         projects_list = []
         for prj in projects:
-            project = {
-                'pk': prj.id,
-                'name': prj.name,
-                'description': prj.description,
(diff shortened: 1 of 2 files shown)


asset_dashboard/views.py
```diff
@@ -24,25 +24,28 @@ def get_context_data(self, *args, **kwargs):
...
-                'total_score': prj.projectscore.total_score,
-                'phase': prj.phase,
-                'zones': list(prj.zones.all().values('name')),
-                'house_districts': list(prj.house_districts.all().values('name')),
-                'senate_districts': list(prj.senate_districts.all().values('name')),
-                'commissioner_districts': list(prj.commissioner_districts.all().values('name'))
-            }
-            projects_list.append(project)
+            phases = Phase.objects.filter(project=prj)
+            for phase in phases:
```
smcalilly, line 32: nested for loop to access the phases for a project :(

smcalilly:
@fgregg nice! changed here: fae9e06

Issue #74, opened 2021-12-06
Does funding need to be a first class object?


fgregg:
Right now, funding sources are basically extended attributes of a project. Should funding be a first class object?

This would be desirable if we wanted two behaviors:

1. If we wanted to associate more than one project with a funding source. Perhaps to see if there was going to be a conflict in funding or whether all the funds related to a funding source were going to be allocated or not?
2. If we wanted to be able to edit a funding source in one place and have that information be updated on every associated project.

fgregg:
It seems like maybe "Sources of Funding" is a first class object, and "Funding" is a relationship between projects and sources?

hancush:
The `FundingSource` (model) and `funding` relationship between `Project` and `FundingSource` makes sense to me.

I think regardless of whether funding can be associated with more than one project, it makes sense to live in its own model / table because it is a conceptually distinct thing from a project, and with #71, we'll be keeping track of a lot more metadata that is unrelated to the project.

It also opens opportunities to create views for managing and tracking progress of funding sources, individually or in the aggregate ("Show me all unallocated grant funding that expires this year") through the dashboard. Right now, AFAICT, you can really only look at funding via a project view.

Issue #72, opened 2021-12-06
What interaction should there be, if any between phase of funding model and phase of project?


fgregg:
Relates to #71

fgregg:
funding is connected to phases, so this is sorted out.

PR #92, opened 2021-12-20
Add Portfolio and PortfolioPhase models

### LINKED ISSUE #85: Create a portfolio model that captures state of portfolio builder
The state of the portfolio builder should be serializable to the portfolio. 

Likely sufficient to this like:
* Name
* Create
* Creation Date
* Updated Date
* One to Many field to Phases 
* An index to indicate the order of phase in the portfolio.

hancush:
## Overview

This PR adds models for persisting a Portfolio and its component Phases. I decided to break `PortfolioPhase` into a related model, rather than defining a `ManyToManyField` between Phase and Portfolio (I assume that a Phase could belong to any number of Portfolios, e.g., if it is not completed in a given year, or if Portfolios are used to compare different scenarios) because we needed somewhere to store a sequence for a Phase within a given Portfolio (which is distinct from its sequence in a `Project`).

This PR also introduces a `SequencedModel` abstract base model for sequencing phases within portfolios and projects. It should ensure that phases are created or updated at the specified sequence, and also update the sequence of downstream phases. I've added a test and docstrings for this functionality, but let me know if it's unclear.

Closes #85

## Testing Instructions

* Confirm that CI passes
* Confirm that a review app is created and released successfully


smcalilly:
@hancush Can you explain what the sequence field does in the `Phase` model? Or point me to an existing discussion about it. I can't find one in the Phase model PR or in any of the issues. That context will help me understand this PR better.

hancush:
@smcalilly Yes, the phase sequence within a project is the order in which it should be completed relative to other phases in the project. So if you have project phases ordered like feasibility, engineering, implementation, the sequence would be 1, 2, and 3 for those phases, respectively.

The sequence within a portfolio is again an order, but it's relative to the other phases in the portfolio. A portfolio might contain the feasibility phase from one project and the construction phase from another, and the portfolio needs a concept of how those phases should be ordered separately from how they're ordered in their respective projects. E.g., CCFP might want to order a construction phase first (sequence 1) in a portfolio, when it's fourth (sequence 4) in its project.

smcalilly (approved):
@hancush Thanks for the explanation. That makes sense, but will the sequencing work whenever they combine funding sources for part of the project or mix and match funding for phases?

For example, what happens if they have partial funding for a design phase and partial funding for an engineering phase, then get more funding for the design phase after the engineering phase has been sequenced? Is that even a possible scenario? I might not be clear on the requirements for this use case nor understand the funding enough.

Code looks good otherwise!

hancush:
@smcalilly That's a great question. 

> what happens if they have partial funding for a design phase and partial funding for an engineering phase, then get more funding for the design phase after the engineering phase has been sequenced?

More than one phase can be ongoing, so I think sequence should refer to the order each phase _starts_. In this case, phase 1 would have sequence 1, and phase 2 would have sequence 2, because phase 1 starts first, even though phase 2 finishes first.

I think that we'll work out the particulars of how to associate funding with phases when we tackle the enriched funding model (#72, #74), but based on what you describe, we ought to consider:

- Allowing multiple funding sources to be associated with a phase
- Associating funding sources with phase within a portfolio, rather than within a project

PR #93, opened 2022-01-06
[wip] asset ui tool

smcalilly:
## Overview
This PR contains all the code we developed for the GIS asset feature. We've been submitting PR's against this branch, so all the code has been reviewed.

Based on this architecture outline that @hancush and I came up with: https://gist.github.com/smcalilly/4d1bc364d7c241b0da1bbab305cd1f98

## Testing Instructions
- visit review app: https://ccfp-asset-d-asset-ui-g1nsd9jw.herokuapp.com/
- test user credentials are named "ccfp-asset-dashboard test user" in the lastpass shared folder
- to use the asset ui: create a phase -> go to the phase -> add an asset


tests/conftest.py
```diff
@@ -182,3 +183,7 @@ def user():
 @pytest.fixture
 def nature_preserves():
     return models.NaturePreserves.objects.create(site_name="Lou Lou's Sanctuary")
+
+@pytest.fixture
+def test_geojson():
```
smcalilly, line 188: went ahead and added this to the fixtures while i had it

tests/conftest.py
```diff
@@ -182,3 +183,7 @@ def user():
...
...","commissioner_district":16,"wastewater":null,"water":null,"ownership":null,"latitude":41.831582689998974,"longitude":-87.83579124652252,"managing_department":null,"improvement_year":null,"addition":null,"fpd_zone":"Central","old_address":null,"street_name_current":"GOLF ROAD","address_number_current":"3300","city_current":null,"zip_city_current":"Brookfield","zip_current":"60513","address_current":"3300 Golf Road, Brookfield, Il 60513","seasonal_closing":null,"pk":"1256"},"geometry":{"type":"Polygon","coordinates":[[[-87.83560191965763,41.831535950500836],[-87.83599323890846,41.831535950500836],[-87.83599674510782,41.83162103513153],[-87.83597896770718,41.83162144389445],[-87.83598268944739,41.83171184553104],[-87.83582143010092,41.83171555722877],[-87.83582343756666,41.831764320755944],[-87.83565389122822,41.83176822409932],[-87.83565112780272,41.83170101300551],[-87.8356330164162,41.831701428500565],[-87.83563262105326,41.83169183553792],[-87.83560304825646,41.83169251970175],[-87.83560057526238,41.831632348936054],[-87.83563540342124,41.83163154758606],[-87.8356316761104,41.831540981256694],[-87.83560215477311,41.831541662899085],[-87.83560191965763,41.831535950500836]]]}}]}')
```
hancush, line 189: Could you just return this as a Python dictionary (formatted with whitespace for readability) instead of JSON loading a string?

asset_dashboard/static/js/components/maps/update.js
```diff
@@ -0,0 +1,132 @@
+import ReactDOM from 'react-dom'
```
hancush, line 1: Can we call this file a noun, such as `clippableMap.js`?

hancush (changes requested):
Thanks for this, @smcalilly! 

At a high level, my thoughts on the Python are that I think it would be easier to set up endpoints to list GIS assets and create local copies with Django REST Framework than to roll our own implementation, as you have here. That would allow you to factor search and save out of your view, leaving it with one responsibility. I opened up a PR with my commit that basically just installs DRF. If you agree that would be a good path, how about that PR comes in next, so the framework is available to you to integrate with in a separate PR?

Left a couple of spare comments around the model and test changes.

The JavaScript looks great! I learned a lot from your mapping docs that translated to understanding your patterns here, which is awesome. One request to rename a component file using a noun.

asset_dashboard/forms.py
```diff
@@ -87,3 +89,50 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+
+class LocalAssetForm:
```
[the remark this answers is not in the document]
hancush: If you agree with going to DRF route, the other comments on this file are moot.

asset_dashboard/models.py
```diff
@@ -250,12 +250,20 @@ class PhaseFundingYear(models.Model):
...
-                                     related_name='assets')
+    # TODO: implement these fields
+    phase = models.ForeignKey('Phase', on_delete=models.CASCADE)
+
+    geom = models.GeometryField(srid=3435)
+
+    # These fields correspond with the unmanaged GIS models.
+    # As far as I know, it's not possible to create a relationship
+    # between two databases with our current setup?
+    building_id = models.BigIntegerField()
```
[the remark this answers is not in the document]
smcalilly: No, there are several asset types we'll need to accept. I was trying to implement the buildings first then add the rest of the tables. This is based off a pattern that Forest and I discussed. We were thinking that we'd have `building_id`, `trail_id`, `poi_id`, `parking_lots_id`, `picnic_grove_id`, etc, with the field holding the key that maps to the different tables in their database. I don't think that was a strict pattern to follow, but it makes sense to have those different fields so that know where the asset comes from. If we used `asset_id` or `quercus_id`, do you have any suggestion on how to keep track of the original table?

asset_dashboard/services.py
```diff
@@ -0,0 +1,68 @@
+import json
```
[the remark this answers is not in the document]
smcalilly: Will DRF be able to cleanly distinguish the different tables that we'll need to search? I thought that having a search service will allow us to have flexibility for searching the geo assets. We'll need to search these tables in their GIS database:
- building
- trails
- points of interest
- parking lots
- picnic grove
- signage

The fields we search aren't consistent between those tables, either. So we'll have to come up with a good way to handle these different cases. For instance, we'll need to search for `fpd_uid` and `name` in the building table, but those don't exist in the trails. For the trails, Garret said we'll need to search by subsystem of the trails. For points of interest, we'll need to filter where `poi_info_id == poin_info_id_group`. 

And re, the services file, I was trying to move to a more organized approach where we could store the different business logic outside of the views, and access what we need in the views, so that we could make the code more organized and maintainable. So ideally the services could be any type of service that we would want to use or reuse in a view. It keeps the view code cleaner by encapsulating that logic that shouldn't live in a view (imo, would be curious to hear other perspectives!)

asset_dashboard/models.py
```diff
@@ -250,12 +250,20 @@ class PhaseFundingYear(models.Model):
...
+class LocalAsset(models.Model):
+    """
+    We save a local copy of a geo asset with this model.
+    """
 
-    name = models.TextField()
-    location = models.GeometryField()
-    project = models.ManyToManyField(Project,
-                                     related_name='assets')
+    # TODO: implement these fields
```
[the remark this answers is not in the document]
smcalilly: Yeah, meant to add that comment above the building_id field. My two comments are related to this:
- https://github.com/fpdcc/ccfp-asset-dashboard/pull/93#discussion_r781454581
- https://github.com/fpdcc/ccfp-asset-dashboard/pull/93#discussion_r781454809

asset_dashboard/forms.py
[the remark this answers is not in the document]
smcalilly: I originally did that, but the django form class have a bunch of html that isn't necessary for saving the assets since we're accepting via ajax and not a standard html form. I had to override those required fields in the init function and it didn't feel right. I was thinking this class was more like a serializer for validating the data, but technically it's like a form too. So it's in this weird limbo -- I think this is where DRF would come in handy?

asset_dashboard/views.py
```diff
@@ -265,6 +270,64 @@ def form_valid(self, form):
...
+        if existing_geometries.exists():
+            context['props'].update({
+                'existing_geoms': serialize('geojson', existing_geometries, geometry_field='geom')
+            })
+
+        if self.request.GET.get('q'):
+            paginated_results, search_results = search_assets(self.request)
+            context['search_results'] = paginated_results
+
+            # Send back the searched geojson, not paginated
```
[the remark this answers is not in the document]
smcalilly: This endpoint is returning a paginated list to show the search results in a table. And it also returns the geojson from the search. I thought the table needed to be paginated so that a user could click through the table. I can implement the pagination in javascript if you think that's redundant/not worth the efficiency to paginate since we're already sending back an entire geojson object? 

This also would be the only reason not to do DRF, since we're serving an html page with a table. If we did do the DRF, then we'd need to serve the list of search results as json and then render the table in react. We'll also be creating another table on this page, the table that lists all of the assets for a project (not yet implemented).

So we need a view that serves the searched assets via both geojson and as a table, plus a list of already saved assets in another table on that page. The view also needs to accept form data (or is that the only endpoint where you're suggesting we use DRF?). With those requirements, is it the appropriate to contain all that logic in one view or separate across multiple views?

Here's the wireframe for this page to give you a better if you haven't seen it: https://app.moqups.com/oyzaWsCDbsZ7xXShLNvzgbpzpBoQ8D1s/edit/page/ae22b960d

tests/conftest.py
smcalilly: I think I saved it like this because the post request is coming in as a json string (?) and I wanted to preserve that for testing. Does that sound right?

smcalilly:
> At a high level, my thoughts on the Python are that I think it would be easier to set up endpoints to list GIS assets and create local copies with Django REST Framework than to roll our own implementation, as you have here.

Just remembered another reason I didn't do pure json endpoints. It's my understanding that for efficiency with a CDN and caching, serving HTML is more efficient? So I was trying to serve the search results the old fashioned way with query string params in the URL rather than some ajax requests that may or may not get the benefit of a CDN?

asset_dashboard/models.py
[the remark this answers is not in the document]
hancush: I think we can push this into another PR, however I'd recommend this article on modeling polymorphism in Django: https://realpython.com/modeling-polymorphism-django-python/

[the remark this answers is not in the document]
hancush: The `django-polymorphic` plug-in is not discussed but is also interesting. https://django-polymorphic.readthedocs.io/en/stable/quickstart.html

asset_dashboard/services.py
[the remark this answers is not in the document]
hancush: I think there's a clean way to be had with DRF, yes. I spiked one idea here: https://github.com/fpdcc/ccfp-asset-dashboard/pull/101

Re: services, it feels a bit code smelly to move code that only one view uses into a separate file. If we write a Django form or an object with a similar API and use a form view, I would expect to find this sort of business logic on the view, either in the `get_context_method` or `form_valid` method.

But let's come back to this discussion once we settle on whether to use DRF or the form pattern.

asset_dashboard/models.py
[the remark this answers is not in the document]
hancush: @smcalilly We talked separately about storing an `asset_id` and `asset_model`.

You can get the pk field of a Django model [like this](https://stackoverflow.com/a/13419091). But it looks like you did something pretty smart with the GIS models, which is creating an `id` field on each GIS model and mapping it under the hood  to the right database column for each asset type. This gives you a unified API for querying on primary key, even though the underlying columns are different.

So, you could write a really simple property on your `LocalAsset` model to get the relevant GIS model instance, if you wanted, no complicated polymorphism required:

```python
@property
def gis_asset(self):
    # good idea to handle invalid model names and asset ids but this is an mvp
    return globals()[self.asset_model].objects.get(id=self.asset_id)
```

(`globals()` is [the current module namespace](https://docs.python.org/3/library/functions.html#globals).)

asset_dashboard/static/js/components/PortfolioTable.js
```diff
@@ -1,7 +1,8 @@
 import React from 'react'
 import ReactTable from './BaseTable'
+import columns from './table-utils/projectColumns'
```
smcalilly, line 3: Import these default column values so we can reuse in both `ProjectsTable` and this `PortfolioTable`. We could put those values in the parent `PortfolioPlanner` container but I thought that would add too much noise to that component, especially since it's already doing so much.

asset_dashboard/static/js/components/BaseTable.js
```diff
@@ -1,51 +1,13 @@
 import React from 'react'
 import { useTable, usePagination, useSortBy } from 'react-table'
 
-const BaseTable = ({ rows = [],  getTrProps = props => props, rowClassNames, selector }) => {
+const BaseTable = ({ rows = [], columns, getTrProps = props => props, rowClassNames, sizeOfPage=15, pageSizeIncrements=[15, 30, 45] }) => {
```
smcalilly, line 4: Made this component more generic so I can use it to show the asset search results. Before this, some details about the `ProjectTable` and `PortfolioTable` were hardcoded in here.

asset_dashboard/static/js/components/BaseTable.js
```diff
@@ -101,11 +63,12 @@ const BaseTable = ({ rows = [],  getTrProps = props => props, rowClassNames, sel
       
       
       <div className="row container" aria-label="Pagination for all potential projects.">
-          <div className="d-flex col justify-content-start align-items-center ml-2">
-            <small className="text-muted">Pages {pageIndex + 1} of {pageOptions.length}</small>
+          <div className="d-flex col align-items-center ml-2">
+            <small className="text-muted">Pages {pageIndex + 1} of {pageOptions.length > 0 ? pageOptions.length : 1 }</small>
```
smcalilly, line 67: fixed this bug that was bothering me. also improved the css for the pagination elements

asset_dashboard/static/js/components/BaseTable.js
```diff
@@ -128,24 +91,26 @@ const BaseTable = ({ rows = [],  getTrProps = props => props, rowClassNames, sel
...
-                key={pSize} 
-                value={pSize} 
-                className="page-item ml-2"  
-                >
-                <button 
-                  onClick={e => {
-                    setPageSize(Number(pSize))
-                  }}
-                  className={`btn btn-light btn-sm ${(pSize == pageSize) ? 'active' : '' }`}
+          {pageSizeIncrements.length > 1 &&
```
smcalilly, line 94: hide the increment buttons if there is only one item in the list (like in the asset search table, since that UI doesn't have much space to take up when considering the layout of the entire page.)

asset_dashboard/static/js/components/SearchInput.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+
+const TextField = styled.input`
```
smcalilly, line 5: need to do better here...

asset_dashboard/static/js/components/map-utils/clipper.js
```diff
@@ -0,0 +1,74 @@
...
+import { debounce } from 'lodash'
+
+// This component controls the interaction for selecting a part, whole, or group of
+// assets. A user can select an area by holding shift and dragging a box
+// over the map with their mouse. This clips the intersection of the bounding box
+// and any geojson assets within the box. The component doesn't return anything in the UI, 
+// but has side effects via a callback that sends the clipped FeatureCollection to 
+// the parent component.
+
+export default function AreaClipper({ geoJson, onClipped }) {
```
smcalilly, line 14: this will likely get thrown out with work for #95 and #97.

asset_dashboard/static/js/components/maps/UpdateAssetsMap.js
```diff
@@ -0,0 +1,217 @@
+import ReactDOM from 'react-dom'
```
smcalilly, line 1: this file is a mess / work in progress. need to clean up before another review.

asset_dashboard/models.py
```diff
@@ -250,12 +250,20 @@ class PhaseFundingYear(models.Model):
...
+    phase = models.ForeignKey('Phase', on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
smcalilly: Will do this when I do the save / delete view.

asset_dashboard/static/js/components/maps/update.js
smcalilly: I renamed it `AssetsMap.js` to match the exported component name. I renamed the component to `AssetsMap` since "update" doesn't completely describe that due to the searched assets showing up on the map, too. I'm starting to name all the files after the exported names, [following the AirBnb style guide](https://github.com/airbnb/javascript#naming--filename-matches-export). And changing the unpythonic snake_case filenames to the ES6 style of camelCase for javascript files that aren't components.

asset_dashboard/static/js/components/BaseTable.js
smcalilly: (actually didn't fix it in the CIP planner tables...hmmm)

tests/conftest.py
smcalilly: Moved into a json file since this is valid json with nulls and can't be formatted in python. I want to save this object for when I write tests.

asset_dashboard/static/js/components/BaseTable.js
smcalilly: also improved the css for the pagination elements

smcalilly: ah, i realized why the code changes weren't showing. i didn't change the paths in the django view and template, so i was served the cached react. ccea10e fixes this.

asset_dashboard/static/js/components/maps/AssetsMap.js
```diff
@@ -0,0 +1,177 @@
...
+            >
+              <AssetTypeOptions />
+            </select>
+            <button 
+              onClick={() => searchAssets()}
+              className='btn btn-warning'>
+                Search
+            </button>
+          </div>
+        </div>
```
hancush: IDK how long queries will be but this feels a bit cramped. What do you think about having this search take up the full width of the container (spanning both the results and map cols)?

Screen Shot 2022-01-20 at 11 28 54 AM

asset_dashboard/static/js/components/maps/AssetsMap.js
```diff
@@ -0,0 +1,177 @@
...
+    fetch(url, {
+      headers: {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json',
+        'X-CSRFTOKEN': Cookies.get('csrftoken')
+      },
+      mode: 'same-origin',
+      method: 'GET'
+    }).then((response) => response.json())
+    .then((data) => setSearchedGeoms(data))
```
hancush, line 92: Can we zoom in on search results easily here?

asset_dashboard/static/js/components/tables/AssetSearchTable.js
```diff
@@ -0,0 +1,36 @@
...
+  const columns = [
+    {
+      Header: 'Identifier',
+      accessor: 'properties.identifier'
+    },
+    {
+      Header: 'Name',
+      accessor: 'properties.name', 
+    }
+  ]
```
hancush, line 16: Can these live in `table_utils/assetSearchColumns.js` for consistency?

asset_dashboard/static/js/components/maps/AssetsMap.js
```diff
@@ -0,0 +1,177 @@
...
+    })
+  }
+
+  function searchAssets() {
+    const url = `/assets/?` + new URLSearchParams({
+      'q': searchText,
+      'asset_type': searchedAssetType
+    })
+
+    fetch(url, {
```
hancush, line 83: This request is kind of heavy when there are many results – is there an easy way to show a loading indicator or even "Loading..." text so the user has some feedback that their search is processing?

hancush (commented):
In general, looking good, @smcalilly, thanks! I left a few questions about UI and spinning off issues, but I don't require any changes at this stage – particularly if they're going to take more than 15-30 minutes.

This PR is getting pretty big. When you're ready to move on to local asset creation, can you create a new branch and PR against this one so we can keep the work and conversation focused? Thanks in advance!

asset_dashboard/static/js/components/maps/AssetsMap.js
smcalilly: Not easily. The first iteration of this worked like that because I was loading the search results from the django template. When the component first loads, the [whenCreated callback executes in the BaseMap component](https://github.com/fpdcc/ccfp-asset-dashboard/blob/4d384825e96be01b749140161b2f8d31ef6be1d7/asset_dashboard/static/js/components/maps/AssetsMap.js#L155) and any geojson that's loaded from props will exist as a layer in the map `whenCreated`. Therefore the map will [zoom to fit those geojson layer](https://github.com/fpdcc/ccfp-asset-dashboard/blob/4d384825e96be01b749140161b2f8d31ef6be1d7/asset_dashboard/static/js/components/maps/AssetsMap.js#L45). 

Fetching the search results from a json api broke the zooming upon search, but it will still work when the page loads and there are existing assets. I tried to fix this but it was a bit more involved, so I decided to get this PR in shape and address that later.

asset_dashboard/static/js/components/maps/AssetsMap.js
smcalilly: What about this? This is less cramped but doesn't span the entire box.

Screen Shot 2022-01-20 at 12 04 23 PM


I wanted to keep the search elements separated from the right side of the page so that the search functionality is compartmentalized on the left side, and the add/edit function is on the right side. To me, that's a clearer user experience.

asset_dashboard/static/js/components/maps/AssetsMap.js
smcalilly: I can just do a simple "Loading..." to start with and we can improve it later.

asset_dashboard/static/js/components/tables/AssetSearchTable.js
smcalilly: Yep.

asset_dashboard/static/js/components/maps/AssetsMap.js
smcalilly: Here's a basic implementation with this commit b397a6. It can definitely be improved but this works for now.
loading
9

asset_dashboard/static/js/components/maps/AssetsMap.js
hancush: Good call, thanks.

asset_dashboard/static/js/components/maps/AssetsMap.js
hancush: Perf, thanks!

asset_dashboard/static/js/components/maps/AssetsMap.js
smcalilly: thanks!

asset_dashboard/static/js/components/maps/AssetsMap.js
smcalilly: might be able to create a bounding box with django-rest-framework-gis? saving here for later reference: https://github.com/openwisp/django-rest-framework-gis#bounding-box-auto_bbox-and-bbox_geo_field

hancush (approved):

Issue #87, opened 2021-12-20
Save the portfolio on UI update. 


fgregg:
If the user makes a change to the portfolio in the UI, save the changes.

PR #98, opened 2022-01-10
Stand up Django REST Framework

hancush:
## Overview

See title. I set this up in service of persisting portfolios (#87), but I think it will be useful to #93 and potentially other features, as well.

## Testing Instructions

* tk

### DIFF
--- a/asset_dashboard/static/js/planner.js
+++ b/asset_dashboard/static/js/planner.js
@@ -75,7 +75,7 @@ class PortfolioPlanner extends React.Component {
     // add the row to the existing portfolio
     const updatedProjectsInPortfolio = [...this.state.portfolio.projects, row]
 
-    const updatedTotals =this.calculateTotals(updatedProjectsInPortfolio)
+    const updatedTotals = this.calculateTotals(updatedProjectsInPortfolio)
 
     // remove the row from the remaining projects
     const updatedRemainingProjects = this.state.remainingProjects.filter((project) => {
--- a/asset_dashboard/endpoints.py
+++ b/asset_dashboard/endpoints.py
@@ -1,0 +1,31 @@
+from django.contrib.auth.models import User
+from rest_framework import viewsets
+
+from asset_dashboard.models import Phase, Portfolio, PortfolioPhase, Project
+from asset_dashboard.serializers import PortfolioSerializer, UserSerializer, \
+    PortfolioPhaseSerializer, PhaseSerializer, ProjectSerializer
+
+
+class PortfolioViewSet(viewsets.ModelViewSet):
+    queryset = Portfolio.objects.all()
+    serializer_class = PortfolioSerializer
+
+
+class UserViewSet(viewsets.ModelViewSet):
+    queryset = User.objects.all()
+    serializer_class = UserSerializer
+
+
+class PortfolioPhaseViewSet(viewsets.ModelViewSet):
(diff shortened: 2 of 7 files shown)


smcalilly (dismissed):
nice! will we want to refactor the views that are using the `django-datatable-view` library?

1) https://github.com/fpdcc/ccfp-asset-dashboard/blob/4dbd6ae6508279882b44cfadf0485577a6868b45/asset_dashboard/views.py#L77
2) https://github.com/fpdcc/ccfp-asset-dashboard/blob/4dbd6ae6508279882b44cfadf0485577a6868b45/asset_dashboard/views.py#L283

hancush:
@smcalilly I think we can keep the data table views the same for now, since the `BaseDataTableView` is doing some extra work for us that I don't want to reimplement in DRF.

asset_dashboard/views.py
```diff
@@ -1,18 +1,22 @@
...
-from django.contrib.auth.mixins import LoginRequiredMixin
+
 from .models import HouseDistrict, Project, ProjectCategory, ProjectScore, \
-    Section, SenateDistrict, CommissionerDistrict, Phase, PhaseFinances
+    Section, SenateDistrict, CommissionerDistrict, Phase, PhaseFinances, Portfolio
 from .forms import ProjectForm, ProjectScoreForm, ProjectCategoryForm, PhaseFinancesForm
-from django.contrib import messages
-from django.db.models import Q
-from django.utils.html import escape
+from .serializers import PortfolioSerializer
```
hancush, line 19: Reordered: Standard library, third-party, internal – alphabetical within groups.

asset_dashboard/static/js/planner.js
```diff
@@ -125,6 +163,62 @@ class PortfolioPlanner extends React.Component {
...
+        'Accept': 'application/json',
+        'Content-Type': 'application/json',
+        'X-CSRFTOKEN': Cookies.get('csrftoken')
+      },
+      body: JSON.stringify(data)
+    }).then(response => {
+      console.log(response.ok, response.status, response.statusText)
+
+      if (!response.ok) {
+        console.error(response.status, response.statusText)
```
smcalilly, line 193: `fetch` has a built-in way of catching errors, i re-learned last week. you could [chain the promises together and ultimately resolve to catch an error if it raises](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#uploading_json_data). i think this would be cleaner and more idiomatic es6. what do you think about that change?

smcalilly (changes requested):
@hancush looks real nice! glad my react code wasn't (seemingly) too hard to work with. i left one suggested change in a comment.

asset_dashboard/static/js/planner.js
hancush: Thanks, @smcalilly! Done in https://github.com/fpdcc/ccfp-asset-dashboard/pull/98/commits/e7c44550ef3555b2b4075deb5fe79da679213847.

hancush:
@smcalilly I have further UI improvements in a separate branch. Once this secures your approval, can I merge into master, then open a new PR, or would you prefer to wait to merge / deploy to staging until the full UI is done?

smcalilly (approved):
Looks good! For merging, it's up to you. We can deal with the merge conflicts that will likely come from my `assets-ui` branch whenever we merge that branch into main.

Issue #94, opened 2022-01-10
Edit assets from phase ui


smcalilly:
The asset selection tool needs to be associated with a phase and accessible through the phase form UI.

Should be able to
- select a phase
- then access the asset tool

smcalilly:
@fgregg Two questions:

1. With this change, do we still need the "Assets" tab in the UI?
![Screen Shot 2022-01-26 at 9 59 03 AM](https://user-images.githubusercontent.com/38969506/151199077-c854d9ab-3eb5-4ccc-a477-81951e775c10.png)

2. Will [the assets list view](https://app.moqups.com/oyzaWsCDbsZ7xXShLNvzgbpzpBoQ8D1s/edit/page/ad64222d5) link from a phase, too? How would a user access that? I'm wondering what the sequence is (phase -> link to the edit tool, then a link to the list view?) Maybe that page isn't even necessary since we have a list of the assets in the asset tool?

fgregg:
i think we don't need the assets tab in the UI. 

i think on the phase view, we might want a map, and then maybe not a separate listing?

PR #110, opened 2022-01-26
Associate the asset tool with a phase

smcalilly:
## Overview
See title. For issue #94. I also changed the LocalAsset.phase relationship to a `ForeignKey` instead of a `ManyToManyField`.

### Demo
phase-assets

### Notes
This PR is still using the prototype's map UX. Those changes are to come with the relevant issues.

## Testing Instructions
- Visit review app: https://ccfp-asset-d-phase-asse-ffhjgx.herokuapp.com/
- Login with the "ccfp-asset-dashboard test user" credentials in the LastPass shared folder.
- Select a project and go to the phases page.
- Select a phase.
- Click "edit assets" in the top left corner.
- Search for some assets. Save them.
- Make sure the assets load.
- Leave the page and come back via the phase edit page. Make sure your saved assets load in the map.


asset_dashboard/views.py
```diff
@@ -275,17 +275,23 @@ class AssetAddEditView(LoginRequiredMixin, TemplateView):
...
         context = super().get_context_data(**kwargs)
 
-        context['project'] = Project.objects.get(id=self.kwargs['pk'])
+        phase = Phase.objects.get(id=self.kwargs['pk'])
+
+        context.update({
+            'phase': phase,
+            'project': Project.objects.filter(phases=phase)[0],
+            'props': {
+                'phase_id': phase.id
```
smcalilly, line 284: need this in `context['props']` so we can access the phase id in the react component.

asset_dashboard/serializers.py
```diff
@@ -92,6 +93,10 @@ def get_geom(self, obj):
 class LocalAssetWriteSerializer(BaseLocalAssetSerializer):
     geom = GeometryField()
 
+    def create(self, validated_data):
+        phase = validated_data.pop('phase')
```
fgregg, line 97: why do you need this? why doesn't

```
LocalAsset.objects.create(**validated_data)
```

not just work

fgregg (changes requested):
looking very clean! do have one question.

asset_dashboard/serializers.py
smcalilly: good question. that was left over from working through this code. it works without overriding the create method, changed here: bf1976c

fgregg (approved):

Issue #112, opened 2022-01-28
improve map zoom


smcalilly:
the map zooms in on existing assets when the map loads the first time. when a user searches, the map stays at that zoom. it would be helpful to use the searched geometries to create a bounding box and zoom

Issue #108, opened 2022-01-26
persist search when a user saves an asset


smcalilly:
[see no. 1 in this PR note](https://github.com/fpdcc/ccfp-asset-dashboard/pull/106#issue-1114332552)

Issue #107, opened 2022-01-26
fix overlapping geometries when searching assets


smcalilly:
[see no. 2 in this PR comment](https://github.com/fpdcc/ccfp-asset-dashboard/pull/106#issue-1114332552)

Issue #97, opened 2022-01-10
Toogle for editing mode vs navigation mode


smcalilly:
Here's an example that lets you toggle different types of interactions: https://geojson.io/#map=3/38.89/-81.21

smcalilly:
This ^ toggle example comes out of the box with the Leaflet.draw plugin. [The react-leaflet-draw library](https://github.com/alex3165/react-leaflet-draw#readme) is the react version of that, and those layer controls are also included. So, I'll introduce that library so that we can have the toggling. Using this library will also give us the ability to draw more dynamic polygons (based on notes from our meeting with Garret, [written down here](https://github.com/fpdcc/ccfp-asset-dashboard/issues/63#issuecomment-1013248556))

fgregg:
sweeeet!

PR #113, opened 2022-01-28
Add leaflet drawing for selecting assets

smcalilly:
## Overview
For issue #97. Adds [`react-leaflet-draw`](https://github.com/alex3165/react-leaflet-draw) for selecting dynamic polygons and rectangles.

### Notes
I had trouble importing the `react-leaflet-draw` library. It seems to be an incompatibility between the latest version of react-leaflet and react-leaflet-draw. I tried using older versions of both libraries based on info I found in some Github issues but this didn't work for me. So I just imported [the `EditControl` component directly from the library's GitHub](https://github.com/alex3165/react-leaflet-draw/blob/master/src/EditControl.js) and it worked with no problems. Weird!

We have these issues for some more map UX improvements:
- #95 
- #96 
- #107 
- #108 
- #109 
- #112

Let me know if you have any other ideas for improvements.

## Testing Instructions
- Visit https://ccfp-asset-d-react-draw-crf578.herokuapp.com
- Login with the "ccfp-asset-dashboard test user" credentials in the LastPass shared folder.
- Select a project and go to the phases page.
- Select a phase.
- Click "edit assets" in the top left corner.
- Search for some assets. 
  - QA the leaflet draw toolbar.
  - Save them.
- Make sure the saved assets load.


asset_dashboard/static/js/components/map_utils/MapClipper.js
```diff
@@ -0,0 +1,110 @@
+import React, { useEffect, useState } from 'react'
+import { FeatureGroup } from 'react-leaflet'
+import clip from 'turf-clip'
+import EditControl from './EditControl'
+
+export default function MapClipper({ geoJson, onClipped }) {
+  const [drawnGeometries, setDrawnGeometries] = useState(null)
```
smcalilly, line 7: ideas for a better variable name? kinda awkward

smcalilly: i had `selectedGeometries` but that was confusing since "selected geometries" could be interpreted as the geometries that were selected from the underlying geojson, and not the geometries created by the draw. even typing that out is confusing.

asset_dashboard/static/js/components/map_utils/MapClipper.js
```diff
@@ -0,0 +1,107 @@
...
+      const bounds = {
+        'type': 'FeatureCollection',
+        'features': Object.values(geometries)
+      }
+  
+      const clippedFeatureCollection = clip(bounds, geoJson)
+  
+      if (clippedFeatureCollection.features.length > 0) {
+        onClipped(clippedFeatureCollection)
+      }
```
hancush, line 76: In what situation are there geometries but no features in the clipped feature collection?

asset_dashboard/static/js/components/map_utils/EditControl.js
```diff
@@ -0,0 +1,148 @@
+// Code source: https://github.com/alex3165/react-leaflet-draw/blob/master/src/EditControl.js
```
hancush, line 1: Can you use a permalink here instead of the master ref, which might change?

hancush (changes requested):
Looking good, @smcalilly. A bit suspicious about the version mismatch, but let's leave that yak to shave another day. Left a few things inline. Thanks for your patience!

asset_dashboard/static/js/components/map_utils/MapClipper.js
smcalilly: if a user draws a geometry but there aren't any underlying geometries.

the `clipGeometries` function will run anytime the `drawnGeometries` state changes, so whenever a user draws, edits, or deletes a polygon/rectangle with the draw tool. if they create a polygon that doesn't intersect with any of the geojson features, then `clippedFeatureCollection.features.length == 0`. in that case, we don't want to call the `onClipped` callback. adding this `clippedFeatureCollection.features.length > 0` is a truthy check so that `onClipped` doesn't get called, otherwise it will send invalid data that results in a 500 error on the server.

asset_dashboard/static/js/components/map_utils/MapClipper.js
```diff
@@ -0,0 +1,107 @@
...
+    clipGeometries(drawnGeometries)
+  }, [drawnGeometries])
+
+  function onEdited(e) {
+    const layerKeys = getLayersKeys(e.layers._layers)
+
+    let editedLayers = {}
+
+    layerKeys.forEach(layerKey => {
+      const layer = e.layers._layers[layerKey]
```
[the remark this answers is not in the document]
smcalilly: nice! changed it here: 7b79e5d

asset_dashboard/static/js/components/map_utils/EditControl.js
```diff
@@ -0,0 +1,148 @@
...
+
```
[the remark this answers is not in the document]
smcalilly: yep, here you go. efb9c07

it was a headache/headscratcher getting it to work with the npm library so i just punted and copy/pasted the code, would love some help if you had time spare/pair.

asset_dashboard/static/js/components/map_utils/EditControl.js
smcalilly: done so here: efb9c07

asset_dashboard/static/js/components/map_utils/MapClipper.js
smcalilly: upon further thought i think this assumption was based on the first iteration of the backend. in the case of the `features` array being empty, we will submit an empty featurecollection to the api and save it in the database (assuming the api would accept that, which i'm guessing would but i haven't tested). so i think we should still protect the backend from accepting an empty feature collection so we don't have a bunch of empty feature geometries stored in the database (assuming it would save an empty geometry...do you know if that is possible?)

hancush: Sounds good. FWIW, I wasn't requesting any changes, I just wanted to understand how this condition is created.

hancush (approved):

PR #115, opened 2022-01-28
Improve UI for creating portfolios

hancush:
## Overview

This PR:

- Adds an explicit button for editing portfolio name
- Prompts the user about unsaved changes if they try to navigate away from the current portfolio
- Adds a portfolio picker to switch portfolios

## Notes

TODO: 
- [x] Default to editable name if no portfolios exist
- [x] Update the portfolio picker when a new portfolio is created.

## Testing Instructions

* Navigate to the review app at https://ccfp-asset-d-hcg-portfo-ratya9.herokuapp.com/cip-planner/ and log in using the test user creds in LastPass
* Confirm a portfolio with one phase is rendered by default
* Try removing the phase, then refreshing the page. Confirm you see a dialog confirming you want to navigate away. Also try clicking links to other pages, etc., and confirm the dialog again appears.
* Create a new portfolio. Try adding/removing phases, saving, etc., and confirm the page state reflects your changes (or lack thereof).
* Refresh the page and confirm the last portfolio you edited is active. (Need to add functionality to update the select with a new option when a portfolio is created, will do Monday.)
* Select the other portfolio from the dropdown and confirm it populates the page correctly.
* Try to break the page in other ways I haven't thought of and let me know what doesn't work. :-)

### DIFF
--- a/asset_dashboard/static/js/planner.js
+++ b/asset_dashboard/static/js/planner.js
@@ -3,6 +3,7 @@ import ReactDOM from 'react-dom'
 import ProjectsTable from './components/ProjectsTable'
 import PortfolioTable from './components/PortfolioTable'
 import PortfolioTotals from './components/PortfolioTotals'
+import PortfolioPicker from './components/PortfolioPicker'
 import SearchInput from './components/FilterComponent'
 import { CSVLink } from 'react-csv'
 import Cookies from 'js-cookie'
@@ -15,6 +16,7 @@ class PortfolioPlanner extends React.Component {
     this.state = {
       allProjects: [],
       remainingProjects: [],
+      allPortfolios: [],
       portfolio: {
         id: null,
         name: '',
@@ -36,6 +38,8 @@ class PortfolioPlanner extends React.Component {
     this.savePortfolioName = this.savePortfolioName.bind(this)
     this.createNewPortfolio = this.createNewPortfolio.bind(this)
     this.alertUserOfUnsavedChanges = this.alertUserOfUnsavedChanges.bind(this)
+    this.selectPortfolio = this.selectPortfolio.bind(this)
+    this.confirmDestroy = this.confirmDestroy.bind(this)
   }
 
   createRegionName(regions) {
@@ -68,12 +72,13 @@ class PortfolioPlanner extends React.Component {
 
     let state = {
       allProjects: projects,
-      remainingProjects: projects
+      remainingProjects: projects,
+      allPortfolios: props.portfolios
     }
 
     // Rehydate state from last edited portfolio, if one exists
-    if (props.portfolio) {
-      const selectedProjectIds = props.portfolio.phases.map(phase => phase.phase)
+    if (props.selectedPortfolio) {
+      const selectedProjectIds = props.selectedPortfolio.phases.map(phase => phase.phase)
 
       const portfolioProjects = projects.filter(
         project => selectedProjectIds.includes(project.key)
@@ -85,8 +90,8 @@ class PortfolioPlanner extends React.Component {
           project => !selectedProjectIds.includes(project.key)
         ),
         portfolio: {
-          id: props.portfolio.id,
-          name: props.portfolio.name,
+          id: props.selectedPortfolio.id,
+          name: props.selectedPortfolio.name,
           projects: portfolioProjects,
           totals: this.calculateTotals(portfolioProjects),
           unsavedChanges: false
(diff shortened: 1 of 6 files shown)


asset_dashboard/static/js/planner.js
```diff
@@ -168,7 +205,7 @@ class PortfolioPlanner extends React.Component {
 
     const data = {
       name: this.state.portfolio.name,
-      user: props.user_id,
```
smcalilly, line 171: props is coming from python and styled as snake case, unless this was styled as camelCase somewhere i'm missing?

asset_dashboard/static/js/planner.js
```diff
@@ -201,20 +238,92 @@ class PortfolioPlanner extends React.Component {
...
   }
 
-  updatePortfolioName(e) {
-    this.setState({
-      portfolio: {...this.state.portfolio, name: e.target.value}
+  savePortfolioName(e) {
+    e.persist()
+
+    return new Promise((resolve, reject) => {
+      const nameInput = document.querySelector('#portfolio-name')
```
smcalilly, line 245: in your callback, you can return the value from the form instead of getting this value from a `document.querySelector`. you should be able to access the form's value from the event: `e.target.value`.

but i just tried getting `e.target.value` from your [`updatePortfolioName` function](https://github.com/fpdcc/ccfp-asset-dashboard/blob/4ff3e767baf2dc6da2718fa1400abec6ac8eafa9/asset_dashboard/static/js/components/PortfolioTable.js#L18) and it's undefined. i think you'd need to add an `onChange` event listener to [your form's input](https://github.com/fpdcc/ccfp-asset-dashboard/blob/4ff3e767baf2dc6da2718fa1400abec6ac8eafa9/asset_dashboard/static/js/components/PortfolioTable.js#L62), get the value from `e.target.value` and set that value to the `PortfolioTable`'s state. then you can grab that value from state in your `updatePortfolioName` function and send it to the `savePortfolioName` callback. 

here's [an example from the search input](https://github.com/fpdcc/ccfp-asset-dashboard/blob/7b1c56859986abacb36be76d35d82fcfe2dcaaff/asset_dashboard/static/js/components/maps/AssetsMap.js#L129) for the asset map. it gets the value from the input change and sets it to state.

smcalilly (changes requested):
This is looking good! Some notes here about the UI and UX, and some suggested code changes in separate comments.

- I used an existing portfolio name for a new portfolio and there was no validation, so it let me save two portfolios with the same name. We could probably validate client side pretty easily, but maybe we could do server side validation? Either way it will add some complexity to the UI code.
- The UI is getting kinda busy with a lot of buttons, likely adding to the cognitive load for a user. I have some ideas, but let me know if you'd want to brainstorm.
  - In the tables, we could style the "View Project" button like a link and not a button. That would clean things up a lot and it would be more semantic since that html element is a link and not a button.
  - Remove the "Portfolio" header from that bootstrap card/container and maybe have it stand alone at the top of the page?
    - Then move the "Edit Name" button to a location that's [closer in proximity](https://www.nngroup.com/articles/gestalt-proximity/) to the portfolio's name. I see that you're following the GitHub pattern, but it's not exactly the same. The GH PR text size is bigger and takes up more real estate on the screen, giving those elements a bit more proximity compared to our portfolio name. Our UI also has the "New Portfolio" button adjacent and breaks the proximity rule. Those two buttons are unrelated in my opinion. It would make more sense to have the "New Portfolio" button next to the "Save Portfolio" button.
  - The "Save Portfolio" button seems out of place at the bottom of that table, especially once they add enough projects to fill that table, making it grow in size. Do you have any ideas where that might be more intuitive? I've been trying to keep all save actions in the top right corner of all the forms for better [predictability](https://uxdworld.com/2021/07/12/why-predictability-is-important-for-ux-design/). Maybe having the "Save Portfolio" next to the "New Portfolio" in the top right corner like you have the "New Portfolio" would make sense, what do you think?

asset_dashboard/static/js/planner.js
hancush: I changed the props to camel case in the view code.

asset_dashboard/static/js/planner.js
hancush: I don't want to update the state until the name is saved, so the user can hit cancel and we don't lose the original name value. I updated this to use FormData(e.target) instead of the query selector.

hancush:
Excellent UI suggestions, @smcalilly! It looks a lot better. This is ready for re-review.

smcalilly (approved):
sick!! works great and looks good.

Issue #114, opened 2022-01-28
add success / failure message after ajax calls in the react components


smcalilly:
- [x] after saving an asset
- [x] after searching for an asset (only failure message)
- [x] after deleting an asset

Issue #95, opened 2022-01-10
improve default behavior for selecting an asset


smcalilly:
If a user selects part of the building, then the entire building should be selected. Not only the part that was selected (how it works now).

smcalilly:
the current code clips based on the intersection of the selected geometry and the geojson that's already on the map.

we only need an intersection if it's a trail (of type `LineString`). otherwise, we need to grab the entire geometry.

PR #117, opened 2022-02-02
Improvements to the asset map UI

smcalilly:
## Overview
For issues:
- #109 
- #108
  - i'm importing [the rooks npm library](https://www.npmjs.com/package/rooks) to use a `useSessionStorage` hook for this. i tried to roll my own (real quick) based on an internet example but it was buggy. let me know if we need to use another library.
- #107 
  - it started to get a bit complicated to parse through the geometries and filter out the intersections with turf js (based on [our comments here](https://github.com/fpdcc/ccfp-asset-dashboard/pull/106#issuecomment-1021648455)), so instead i styled the search geometries to have a dashed line and the existing geometries to have a solid line. let me know if that looks awkward or you have any suggestions to tweak the styling.
  - note that this was largely an issue with trails (of geometry type `LineString`)
- #112 
  - i made a `MapZoom` component that uses [the react-leaflet `useMap` hook](https://react-leaflet.js.org/docs/api-map/#usemap). the component zooms to the search geometries if there are any. if there aren't but there are existing geometries, then it will zoom to those geometries.
- renamed some state objects to present tense

### Notes
Remaining UI improvements not in this PR:
- #95
  - i'm still figuring out how to parse out the different types of geometries and leverage turf js. unfortunately, [the turf intersection](https://turfjs.org/docs/#intersect) only accepts polygons/multipolygons and not feature collections (which are the geometry types when we need to use turf). this is requires a bit heavier lifting that i haven't ironed out and could use more scrutiny in a standalone pr.
- #114
  - to do

## Testing Instructions
- visit review app: https://ccfp-asset-d-existing-a-qfqdlr.herokuapp.com
- login with lastpass "ccfp-asset-dashboard test user" credentials
- #109 
 - test the delete button and make sure it deletes your assets from the map from the table
- #108
  - search for assets and then reload by refreshing the page. you can also save an asset or delete one, and your search should persist after the page reloads.
- #107 
  - search for some trails and save a portion trails. when the map reloads (it should've persisted the search!), you should be able to distinguish between the trails you saved and the trails from the search.
- #112 
  - all of the searching in your tests should zoom so that the map fits the bounds of your search geometries. 
  - to test that it zooms to existing geometries on a fresh page load, open the web page in a new tab, since the search geometries are stored in session storage.

### DIFF
--- a/asset_dashboard/static/js/components/map_utils/MapZoom.js
+++ b/asset_dashboard/static/js/components/map_utils/MapZoom.js
@@ -1,0 +1,17 @@
+import React, { useEffect } from 'react'
+import { useMap } from 'react-leaflet'
+import zoomToSearchGeometries from './zoomToSearchGeometries'
+
+
+function MapZoom({ searchGeoms }) { 
+  const map = useMap()
+
+  useEffect(() => {
+    const group = new L.featureGroup()
+    zoomToSearchGeometries(map, group)
+  }, [searchGeoms])
+
+  return null
+}
+
+export default MapZoom
--- a/asset_dashboard/static/js/components/map_utils/zoomToSearchGeometries.js
+++ b/asset_dashboard/static/js/components/map_utils/zoomToSearchGeometries.js
@@ -1,0 +1,17 @@
+export default function zoomToSearchGeometries(map, group) {
+    map.eachLayer((layer) => {
+      if (layer.feature?.properties.identifier) {
+        // If "identifier" is in "properties",
+        // then this is a feature from the search.
+        // This is extremely dependent on the search geometries
+        // having this "identifier" key, so beware!
+        // I haven't been able to figure out a better
+        // way of zooming in on the search results.
+        group.addLayer(layer)
+      }
+    })
+  
+    if (Object.keys(group._layers).length > 0) {
+      map.fitBounds(group.getBounds())
+    }
+  }
--- a/asset_dashboard/static/js/components/map_utils/zoomToSearchGeometries.js
+++ b/asset_dashboard/static/js/components/map_utils/zoomToSearchGeometries.js
@@ -1,12 +1,6 @@
 export default function zoomToSearchGeometries(map, group) {
     map.eachLayer((layer) => {
-      if (layer.feature?.properties.identifier) {
-        // If "identifier" is in "properties",
-        // then this is a feature from the search.
-        // This is extremely dependent on the search geometries
-        // having this "identifier" key, so beware!
-        // I haven't been able to figure out a better
-        // way of zooming in on the search results.
+      if (layer.feature?.properties.source === 'search') {
         group.addLayer(layer)
       }
     })
--- a/asset_dashboard/static/js/components/maps/AssetsMap.js
+++ b/asset_dashboard/static/js/components/maps/AssetsMap.js
@@ -3,11 +3,11 @@ import React, { useState, useEffect } from 'react'
 import { GeoJSON } from 'react-leaflet'
 import hash from 'object-hash'
 import Cookies from 'js-cookie'
+import { useSessionstorageState } from 'rooks'
 import BaseMap from './BaseMap'
 import AssetSearchTable from '../tables/AssetSearchTable'
 import ExistingAssetsTable from '../tables/ExistingAssetsTable'
 import MapClipper from '../map_utils/MapClipper'
-import { useSessionstorageState } from 'rooks'
 
 function AssetTypeOptions() {
   // these options could come from the server but hardcoding for now 
--- a/asset_dashboard/static/js/components/maps/AssetsMap.js
+++ b/asset_dashboard/static/js/components/maps/AssetsMap.js
@@ -7,6 +7,7 @@ import BaseMap from './BaseMap'
 import AssetSearchTable from '../tables/AssetSearchTable'
 import ExistingAssetsTable from '../tables/ExistingAssetsTable'
 import MapClipper from '../map_utils/MapClipper'
+import { useSessionstorageState } from 'rooks'
 
 function AssetTypeOptions() {
   // these options could come from the server but hardcoding for now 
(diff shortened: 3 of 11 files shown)


asset_dashboard/static/js/components/map_utils/MapZoom.js
```diff
@@ -0,0 +1,17 @@
...
+import zoomToSearchGeometries from './zoomToSearchGeometries'
+
+
+function MapZoom({ searchGeoms }) { 
+  const map = useMap()
+
+  useEffect(() => {
+    const group = new L.featureGroup()
+    zoomToSearchGeometries(map, group)
+  }, [searchGeoms])
```
smcalilly, line 12: this will run whenever the incoming `searchGeoms` change

hancush: Woah, cool trick!

asset_dashboard/static/js/components/map_utils/zoomToSearchGeometries.js
```diff
@@ -0,0 +1,17 @@
+export default function zoomToSearchGeometries(map, group) {
+    map.eachLayer((layer) => {
+      if (layer.feature?.properties.identifier) {
+        // If "identifier" is in "properties",
+        // then this is a feature from the search.
+        // This is extremely dependent on the search geometries
+        // having this "identifier" key, so beware!
+        // I haven't been able to figure out a better
+        // way of zooming in on the search results.
```
hancush, line 9: Hmmmm, is there a more explicit way to identify search results? Like, adding a method attribute like "source: search" or something to features returned from the API? That feels hacky, too... what do you think?

asset_dashboard/static/js/components/maps/AssetsMap.js
```diff
@@ -118,83 +127,93 @@ function AssetsMap(props) {
...
+              zoom={11}
+              whenCreated={onMapCreated}>
+              {searchGeoms && 
+                <>
+                  <GeoJSON 
+                    data={searchGeoms} 
+                    // Hash key tells the geojson to re-render 
+                    // when the state changes: https://stackoverflow.com/a/46593710
+                    key={hash(searchGeoms)} 
+                    style={{color: 'black', dashArray: '5,10', weight: '0.75'}}
```
hancush, line 195: I really like the contrast between results and selections. Good choice!

asset_dashboard/static/js/components/tables/ExistingAssetsTable.js
```diff
@@ -0,0 +1,48 @@
...
+        'Content-Type': 'application/json',
+        'X-CSRFTOKEN': Cookies.get('csrftoken')
+      },
+      method: 'DELETE',
+      mode: 'same-origin'
+    }).then((response) => {
+      console.log('response', response)
+      if (response.status == 204) {
+        // TODO: this reloads the page but clears the user search...
+        // Need to come up with way to reload by rehydrating the previous state
```
hancush, line 21: Is this true anymore? Seems to be working fine for me, and I would expect the session storage approach to retain search stuff on reload!

hancush (changes requested):
This looks great and works as advertised, @smcalilly. I left two questions for you to respond to inline. Changes not necessarily required, but want to hear your thoughts before deciding / approving.

And an odd UI question: What is the function of the table of search results? It doesn't seem to be adding much at the moment. Are there plans to add something like, highlight the building on the map when a result is hovered and select it when a result is clicked?

smcalilly:
> And an odd UI question: What is the function of the table of search results? It doesn't seem to be adding much at the moment. Are there plans to add something like, highlight the building on the map when a result is hovered and select it when a result is clicked?

This is a good question and I agree, the table seems pointless as is. I've been wanting to add that feature you suggest and was planning to bring it up. I don't think it would be too difficult, but we'd be nearing the limit of complexity with this code.

Somewhat related, we could also go the opposite way, where a user selects a geometry on the map and allows them to save it by simply clicking. And it could also highlight in the table.

Garret has said that their staff knows the id's since it's common for them to use it, so I think having an identifier can be helpful. We could also add a popup easily that shows this info, like 15 minutes easy.

What do you think? @hancush, cc @fgregg 

To recap, some ways to improve it:
1. a user could click on the table and it takes them to that asset on the map. the click would also load the asset so it can be saved.
2. a user could click on an asset in the map and they could save it. the click would also highlight the asset in the table.
   - alternatively, we could have a very simple implementation of a popup with info about each assest.

asset_dashboard/static/js/components/tables/ExistingAssetsTable.js
smcalilly: no, not true anymore. i've deleted this in the code that i've finished and about to open a PR for issue #114

asset_dashboard/static/js/components/map_utils/zoomToSearchGeometries.js
smcalilly: a method attribute in the geojson response? would this be added to the serializer?

smcalilly: here's a sample of geojson from the search:

```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [-87.87729052716881, 41.70864290520196],
          [-87.87722544938457, 41.708603704430736],
          [-87.87716885402861, 41.708534150345045],
          [-87.87713091829738, 41.70845537603724]
        ]
      },
      "properties": {
        "identifier": 3273,
        "name": "Little Red Schoolhouse Nature Center Trails"
      }
    },
  ]
}
```

are are you suggesting we put it in the properties or in a higher level, like
```
"type": "FeatureCollection",
"source": "search",
  "features":
  ...
```

i think that would work, though not entirely how with drf. but i'm sure we could figure it out.

hancush: > a method attribute in the properties of the geojson response? would this be added to the serializer?

Yes, exactly, so it would show up as attribute in the properties of each feature. It's just one idea, open to others.

hancush:
@smcalilly I think you should spin off the table UI into a separate issue.

asset_dashboard/static/js/components/map_utils/zoomToSearchGeometries.js
smcalilly: yeah i think that's better than what i'm doing here because it's more explicit without that big code comment.

smcalilly: @hancush like so? 

i made a [source serializer class](https://github.com/fpdcc/ccfp-asset-dashboard/blob/99dfc421b4e0f98e90fd86397aa7c88e83c706e6/asset_dashboard/serializers.py#L101) for [overriding when we need that field](https://github.com/fpdcc/ccfp-asset-dashboard/blob/99dfc421b4e0f98e90fd86397aa7c88e83c706e6/asset_dashboard/serializers.py#L110). and that helped [clean up the javascript](https://github.com/fpdcc/ccfp-asset-dashboard/blob/99dfc421b4e0f98e90fd86397aa7c88e83c706e6/asset_dashboard/static/js/components/map_utils/zoomToSearchGeometries.js#L3). (sorry these pieces point at links and not one commit, i committed and then realized i need to reuse the code, so refactored)

hancush (approved):
👏

PR #122, opened 2022-02-04
Add success and error messages to the ajax responses

smcalilly:
## Overview

Closes:
- #114 

### Notes
We might want to improve the success messages when the page reloads. The message shows up and the page reloads, which kinda seems like the action is done, but the page is still reloading and we don't want users to use the page during that. Maybe adding the success message near the "save asset" button, plus giving a loading indicator would be more helpful...thoughts?

## Testing Instructions
- review app: https://ccfp-asset-d-ajax-messa-ukywag.herokuapp.com
- test credentials are called "ccfp-asset-dashboard test user"
- search. turn off your wifi network and see an error message.
- select/save an asset. see success message. select/save again with network turned off, see error message.
- delete an asset. see messages etc

### DIFF
--- a/asset_dashboard/static/js/components/maps/SelectAssetsMap.js
+++ b/asset_dashboard/static/js/components/maps/SelectAssetsMap.js
@@ -33,6 +33,7 @@ function SelectAssetsMap(props) {
   const [searchAssetType, setSearchAssetType] = useSessionstorageState('searchAssetTypes', 'buildings')
   const [isLoading, setIsLoading] = useState(false)
   const [phaseId, setPhaseId] = useState(null)
+  const [helpText, setHelpText] = useState(null)
 
   useEffect(() => {
     if (props?.existing_assets) {
@@ -80,7 +81,7 @@ function SelectAssetsMap(props) {
         body: JSON.stringify(data)
     }).then((response) => {
       if (response.status == 201) {
-        renderMessage('Assets successfully saved.', 'success')
+        setHelpText('Assets successfully saved.')
         location.reload()
       }
     }).catch(error => {
@@ -109,6 +110,7 @@ function SelectAssetsMap(props) {
     .then((data) => {
       setIsLoading(false)
       setSearchGeoms(data)
+      // setHelpText('Use the map toolbar to select an asset.')
     })
     .catch(error => {
       setIsLoading(false)
@@ -163,12 +165,11 @@ function SelectAssetsMap(props) {
             {clippedGeoms 
               ?
                 <button 
-                  className='btn btn-primary'
+                  className={helpText ? 'btn btn-secondary' : 'btn btn-primary'}
                   onClick={() => saveGeometries()}>
-                  Save Asset
+                  {helpText ? helpText : 'Save Assets'}
                 </button>
-              :
-                <p>Use the map toolbar to select an asset.</p>
+              : <p>Use the map toolbar to select assets.</p>
             }
           </div>
           <div className='map-container' aria-label='Asset Selection Map'>
--- a/asset_dashboard/static/js/components/maps/SelectAssetsMap.js
+++ b/asset_dashboard/static/js/components/maps/SelectAssetsMap.js
@@ -11,7 +11,7 @@ import MapClipper from '../map_utils/MapClipper'
 import MapZoom from '../map_utils/MapZoom'
 import zoomToSearchGeometries from '../map_utils/zoomToSearchGeometries'
 import zoomToExistingGeometries from '../map_utils/zoomToExistingGeometries'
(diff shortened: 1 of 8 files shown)


hancush (commented):
Agree it's pretty confusing for the success message to disappear when the page reloads. What if you saved that message to the session and check for messages when the page loads, clearing them from the session after they're rendered?

smcalilly:
> What if you saved that message to the session and check for messages when the page loads, clearing them from the session after they're rendered?

@hancush do you know of any DataMade examples that do this? I'm having a hard time finding the best way to do this. [This is the best I can find in SO](https://stackoverflow.com/questions/32155851/session-as-persistent-storage-in-django-rest). I'm not sure the best way to access the session in a [generic view](https://www.django-rest-framework.org/api-guide/generic-views/)?

Perhaps for successful responses, we can do something similar to what you're doing in the CIP planner, instead of using a message-like thing in the template? Like how you're showing that the portfolio saved where the button is. And then for error messages, we can show the message in the Django template, since that seems natural and isn't dependent on a page reload.

hancush:
That would work, @smcalilly! I like the pragmatism.

smcalilly:
@hancush I made these changes:
- after saving assets, show success message in the button
- after deleting assets, show a success message above the table. show an error message above that table too — the previous code showed the message at the top of the page and i thought that was not helpful because you can't see it unless you scroll up.

asset_dashboard/static/js/components/maps/SelectAssetsMap.js
```diff
@@ -79,14 +81,11 @@ function SelectAssetsMap(props) {
...
-        // TODO: this reloads the page but clears the user search...
-        // Need to come up with way to reload by rehydrating the previous state
+        setAjaxMessage('Assets successfully saved.')
         location.reload()
-        // TODO: show success message
-        // https://stackoverflow.com/questions/13256817/django-how-to-show-messages-under-ajax-function
       }
     }).catch(error => {
-      // TODO show error message
+      renderMessage('An error occurred saving the selected assets. Please try again.', 'danger')
```
hancush, line 88: The `renderMessage` function seems to do mostly the same thing as the AJAX message state approach. Is there a way to be consistent here, instead of doing the same thing two ways?

asset_dashboard/static/js/components/maps/SelectAssetsMap.js
```diff
@@ -79,14 +81,11 @@ function SelectAssetsMap(props) {
         body: JSON.stringify(data)
     }).then((response) => {
       if (response.status == 201) {
-        // TODO: this reloads the page but clears the user search...
-        // Need to come up with way to reload by rehydrating the previous state
+        setAjaxMessage('Assets successfully saved.')
```
hancush, line 84: We talked about not having the page reload after rendering a message. I understand we're using the page reload to update the local asset list serverside. What about using the browser session state via the rook library, like you did to persist the user's last search, to store a success message, then render and remove it from the state when the page reloads?

hancush (changes requested):
A few things inline!

asset_dashboard/static/js/components/maps/SelectAssetsMap.js
smcalilly: @hancush Ah, I guess I didn't explain what I was talking about well enough / we we're thinking two different things. I didn't like how the message persisted **after** the page reloads. I thought it makes sense that the message shows up after the ajax call, and then the page reloads, and then the message is gone once the page reloads, like how it works now. 

I can use session storage, but I think that might add some extra complexity to clear it after a page reload. I'll have to check that the message is there every time the component mounts, and then show it, and then clear it from sessionStorage once the user exits out of it.

smcalilly: I'm seeing now that the reload is a lot quicker in the deployment so my use case doesn't make as much sense. It takes a long time to reload in development so it seemed like that was enough time to show the message. But I think that you're right and session storage messages would be the best way to go.

I thought when you were saying "save that message to the session" you meant [the django session](https://docs.djangoproject.com/en/4.0/topics/http/sessions/) and not session storage.

hancush: I wondered if the session mixup was the case, @smcalilly! 😓 Have what you need to make that change?

smcalilly:
@hancush I made the requested changes. I'm not sure how to render the message only after it reloads since adding that to state will show the component. Do you have any ideas, or is this ok as is?

Add and delete some assets to see it in action: https://ccfp-asset-d-ajax-messa-ukywag.herokuapp.com/projects/phases/edit/4/assets/

Issue #111, opened 2022-01-26
add a map with a phase's geo assets to the phase page


smcalilly:
[based on a discussion](https://github.com/fpdcc/ccfp-asset-dashboard/issues/94#issuecomment-1022343600) in issue #94.

PR #124, opened 2022-02-04
Add asset map to phase edit page

smcalilly:
## Overview

Closes:
- #111

Also:
- factored out the zoom code from the original map into a function so we can use that logic in this map
- changed some variable and file name changes for the other map
- a CSS name change because my previous one sounded awkward

## Notes
Let me know if you have any UI improvement ideas...not sure about this one.

## Testing Instructions
- visit https://ccfp-asset-d-phase-map-lfgp5z5.herokuapp.com
- login with `ccfp-asset-dashboard test user` credentials in lastpass
- visit a phase's edit page. add some assets via the button on that page.
- go back to the phase edit page. confirm that your assets show up in the map.


hancush (commented):
Code looks good! Let's talk about UI in our internal check in.

PR #128, opened 2022-02-15
Add search and save for the remaining tables

### LINKED ISSUE #118: Add remaining tables to the asset search
I asked Garret what tables we need to search, and on what fields.

- [x] `buildings` table
  - search by `fpd_uid` and `building_name`
- [x] `trails`
    - "a trail subsystem will get them close enough to where they need to find what they're looking for" (rough quote)
- [x] `pointsofinterest`
  - similar to buildings, search for name
  - filter on `poi_info_id == poi_info_id_group`
- [x] `picnicgroves`
  - this table has a `fpd_uid`, but we need to double check that's the right field for searching
- [x] `parking_lots`
  - need to get clarification about what fields to search
- [ ] `signage`
  - this will require some filtering because it's a catch all for different types of points (like benches, etc) -- need clarification
  
Here's a link to their quercus diagram for reference: https://drive.google.com/file/d/1rqvbMoNbCuAxBpXKhwUa1WbEsdGQVYoZ/view?usp=sharing

_Originally posted by @smcalilly in https://github.com/fpdcc/ccfp-asset-dashboard/issues/62#issuecomment-1013240850_

smcalilly:
## Overview
Adds search for the remaining tables. Closes #118.

Other changes include:
- DB migration for `LocalAsset` model to accommodate for when the `asset_id` is a charfield instead of an integer. This is required because the `PicnicGroves` fpd_id is a charfield.
- Another DB migration for `LocalAsset` model to allow null and blank `asset_id` fields. This field was meant to save the `fpd_uid` but in testing this PR, I found an edge case where sometimes that field is null.
  - Also had to create a `NullableCharField` in the `BaseLocalAssetSerializer` so we could save this field.
  - The discovery of this edge case led to a discovery of improper 400 error handling when submitting a post request via ajax, so I added better error handling for that.
- Added a popup to the asset geometries in the map.
- For the `ParkingLots` model, changed the table to search the `"acer"."parking_lots_union_mv"` and changed the fields to match that materialized view.

### Notes
This searches all of the tables that we've discussed with Garret. If we need to fine tune the columns we search, it will be as simple as adding the desired columns to @hancush's clever [`Search` subclass](https://github.com/fpdcc/ccfp-asset-dashboard/blob/757805367f3452f8de4ec49be908084011e11e3d/asset_dashboard/models.py#L381) solution. That subclass works in unison with [the `AssetViewSet's `get_queryset` method](https://github.com/fpdcc/ccfp-asset-dashboard/blob/757805367f3452f8de4ec49be908084011e11e3d/asset_dashboard/endpoints.py#L85).

## Testing Instructions
- visit review app: https://ccfp-asset-d-remaining--jg5dia.herokuapp.com/
- login with the `ccfp-asset-dashboard test user` credentials in lastpass
- go to the asset map tool
- search and save these types of assets:
  - Points of Interest
  - Picnic Groves
  - Parking Lots
  - Signage
- you can test the NullableCharField in the `BaseLocalAssetSerializer` by searching those asset types with the text "red". that search should return some assets where that field null. you should be able to save the assets when they have a identifier field.


asset_dashboard/models.py
```diff
@@ -700,39 +700,36 @@ class Meta(GISModel.Meta):
...
 
-
 class ParkingLots(GISModel):
-    """Parking lot polygons, for all public and non-public lots."""
-
     class Meta(GISModel.Meta):
-        db_table = '"quercus"."parking_lots"'
-
-    id = models.AutoField(primary_key=True, db_column='parking_lots_id')
+        db_table = '"acer"."parking_lots_union_mv"'
```
smcalilly, line 705: query this materialized view in the `acer` table instead of the `quercus` table.

asset_dashboard/migrations/0026_auto_20220207_2145.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('asset_dashboard', '0025_auto_20220126_1731'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='localasset',
+            name='asset_id',
+            field=models.CharField(max_length=1000000),
```
fgregg, line 16: if the charfield is that long, we should use a textfield

asset_dashboard/migrations/0027_auto_20220215_1707.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('asset_dashboard', '0026_auto_20220207_2145'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='localasset',
+            name='asset_id',
+            field=models.CharField(blank=True, max_length=1000000, null=True),
```
fgregg, line 16: ditto on textfield

asset_dashboard/models.py
```diff
@@ -259,7 +259,7 @@ class LocalAsset(models.Model):
 
     geom = models.GeometryField(srid=3435)
 
-    asset_id = models.IntegerField()
+    asset_id = models.CharField(max_length=1000000, null=True, blank=True)
```
fgregg, line 262: text field

asset_dashboard/models.py
```diff
@@ -700,39 +700,36 @@ class Meta(GISModel.Meta):
     remarks = models.TextField()
     photos = models.TextField()
 
-
 class ParkingLots(GISModel):
-    """Parking lot polygons, for all public and non-public lots."""
```
fgregg, line 705: why did we lose this doc string?

fgregg (changes requested):
looks good!

asset_dashboard/models.py
smcalilly: oops! added back here: 2363933

asset_dashboard/models.py
smcalilly: changed here: b1f7117

smcalilly:
@fgregg this is ready for re-review. you can test the review app at this new instance: https://ccfp-asset-d-remaining--jg5dia.herokuapp.com/

fgregg (approved):
looks good @smcalilly

Issue #75, opened 2021-12-06
Design interface for richer funding model.


fgregg:
Depends on #74

Issue #71, opened 2021-12-06
Enrich Funding models to capture many new attributes


fgregg:
* Phase
* Funded/unfunded
* Funding source
* Year takes place
* Project category
* Funding amount
* Highlight total budget from all phases.
* Grant and Grant amount for phase
* Grant obligated w/ completion date of obligation
* Completed – Yes/No
* Required – Yes/No
* Estimated Bid Quarter
* Notes

fgregg:
We need to interview @glw about the details of this.

fgregg:
- Phase -> Money for phase -> Funding source

Money for phase
- amount
- estimated bid quarter
- obligated completion date

Funding Source
- Type
  - grant
  - other things

Phase 
- what are completed 
- **and required**
- year takes place 
- funded / unfunded

PR #133, opened 2022-03-08
Improve funding models

smcalilly:
## Overview

Includes changes to the models related to phase and funding. Adds a form for the new `FundingStream` model. Adds more information about funding to the phase detail page.

For issues:
- #71 
- #74 
- #75 

## Demo
Screen Shot 2022-03-09 at 9 41 25 AM
Screen Shot 2022-03-09 at 9 41 34 AM

### Notes
The UI for the funding form is especially awkward because it doesn't have many fields and spans across the entire page (see second image above). We could maybe make that a modal form on the phase form page, do you have any thoughts @fgregg?

## Testing Instructions
- visit review app: https://ccfp-asset-d-funding-mo-4urajt.herokuapp.com/
- sign in with the `ccfp-asset-dashboard test user` credentials in last pass
- go to a phase. 
- add funding. make sure it shows up as expected after creating the funding instance.
- edit the funding and make sure it reloads the correct data.
- add another funding stream and make sure it renders in the table.
- go back to the phase list view for the project and make sure a total of all your funding shows up in the table.


asset_dashboard/models.py
```diff
@@ -221,33 +237,26 @@ class Meta:
...
+        ('rollover', 'Rollover')
     ]
 
-    phase = models.OneToOneField('Phase', on_delete=models.CASCADE)
     budget = MoneyField(default_currency='USD',
                         default=0.00,
                         max_digits=11)
+    obligated_year = models.IntegerField(null=True, blank=True)
+    # TODO: better model field type? do we need months/days?
+    obligated_completion_date = models.IntegerField(null=True, blank=True)
```
smcalilly, line 254: Do we need months/days or does an IntegerField work here (so we only save year)?

fgregg:
i think it's find to start with a wide form (they like spreadsheet-like interfaces)

Issue #126, opened 2022-02-09
Programatically get a project's zones and political districts based on the project's GIS assets


smcalilly:
Will also need to remove this from the project edit form and instead show a list of zones/districts

smcalilly:
We need to get the boundaries of these zones

smcalilly:
There are several pieces to do this. At a high level:
- Add a geometry field to the district and zone models to have the boundaries
- Write import script to associate the boundaries from their GIS database to our models
- Add attribute to the project model that calculates the districts/zones for a project
- List the districts/zones based on the project model attribute

At a low level:

## 1. Add geometry field to these models
- [ ] HouseDistrict
- [ ] SenateDistrict
- [ ] CommissionerDistrict
- [ ] Zone

Alternatively, we could have a new Boundary model with a geometry and then relate each model to it.

## 2. Import boundaries from their GIS database
Their `pinus` has these tables:
- [ ] il_house_cb_2017_17_sldl_500k_cc
- [ ] il_senate_cb_2017_17_sldu_500k_cc
- [ ] commdistricts
- [ ] zones

One problem with these data: the house and senate districts are a specific year, so this will break whenever new house/senate boundaries are drawn and a new table is added with the new boundaries.

## 3. Add attribute to Project model for getting the boundaries
I'm not sure the best approach here. Using a phase's GIS assets, we'll need to find the intersection of each asset with each district/zone boundary.
- Do we want an attribute that calculates the boundaries on the fly (meaning it gets the boundaries each time a Project is queried)? This could be inefficient.
- Alternatively, we could somehow save the districts/zones whenever a GIS asset is saved.

## List the districts/zones in the Project detail view/UI
We'll need to figure out our strategy for step 3 before we can do this. Once we figure out that, this piece should be straightfoward.

fgregg:
sounds good.

write the import_shape script assuming that we have the data in the repo somewhere as geojson already.

PR #143, opened 2022-03-23
calculate political boundaries when an asset is saved + incorporate feature in project overview ui

smcalilly:
## Overview

For:
- #126 

#### How do we calculate this for production?
All of the projects on the production database already have districts associated with them. New relationships won't be created until they save some assets. Do we delete these existing relationships, or what?

I'm using a new text editor and it appears that I have a plugin installed that removed indentations on lines with no code. Sorry for the extra diff!

## Testing Instructions
- visit review app: https://ccfp-asset-d-feature-pr-awf7kf.herokuapp.com
- login with the `ccfp-asset-dashboard test user` credentials in last pass
- view a project and edit/add a phase. save some assets.
- revisit the project detail page and make sure the districts/zones were updated.


fgregg:
> The house and senate district boundaries in their DB are dated 2017 and already out of date. Perhaps, regarding the geojson comment, you were thinking that we would download the latest political boundaries as geojson? If we want to do that, the import script will be easily adaptable.

yeah, it doesn't really make sense to have their DB to be the source of truth, and it's pretty easy to store these things statically

fgregg:
i think it's fine to do it by hand, as long as you document in the README where you got them. for the zones, let's get the data from their database, but then save it as a geojson file so it's consistent with the others.

smcalilly:
I got the geo data from public sources. I went ahead and wrote a makefile for the import script, it didn't take much time.

fgregg (approved):
this is super clean!

Issue #134, opened 2022-03-17
filter buildings to only show FPD owned buildings


glw:
Building assets should be filtered to only include FPD owned buildings.

select * from quercus.buildings where ownership = 'fpdcc';

glw:
oh and demolished = 'no'

PR #164, opened 2022-03-30
Add improvements to the GIS search

smcalilly:
## Overview
Closes:
- #151 
- #152 
- #150
- #134 
- fixed a bug I found where a `Point` geometry type would cause the map to crash when clicked on

### Notes
For $150 and #151, I added some javascript code to dynamically render the table columns based on the incoming geojson. It's a bit hacky but it works. The geojson serializers fields are standardized across the different types of assets. If they want to keep adding attributes to the different asset types, then we might want to change our approach and make something more flexible, otherwise this could become messy and unwieldy to deal with.

Also, a thought I had while adding a new attribute and filter to the picnic grove... We could improve the search results by adding a dropdown selection in the UI that allows them what field they want to search on. This would take some work with more significant changes but would be a nice thing to have.

## Testing Instructions
- visit https://ccfp-asset-d-feature-gi-dl0bti.herokuapp.com
- credentials: `ccfp-asset-dashboard test user`
- search for buildings and picnic groves
  - search by these fields:
    - building: `complex` (like camp dan beard)
    - picnic grove: `grove`
  - make sure the new fields show up in the table and popup
- search other asset types and make sure they work correctly
- save some assets and make sure it still works


asset_dashboard/endpoints.py
```diff
@@ -64,7 +64,7 @@ def model_cls(self):
             'trails': {'model': TrailsInfo},
             'points_of_interest': {'model': PoiInfo},
             'picnic_groves': {'model': PicnicGroves},
-            'parking_lots': {'model': PoiInfo, 'check_for_not_null': True}
```
smcalilly, line 67: removed this because it's an `AND` query and we needed to add some more `AND`s to other models. so i made some additions to the model.Search subclass. see related changes in this endpoints.py file and the models.py file.

asset_dashboard/static/js/components/map_utils/ShowPopup.js
```diff
@@ -10,12 +10,14 @@ export default function ShowPopup({ geojson }) {
         layer.openPopup()
 
         // Make the layer appear selected.
-        layer.setStyle({
-          fillColor: '#3388ff',
-          fillOpacity: '0.2',
-          color: '#3388ff',
-          weight: '4'
-        })
+        if (layer.setStyle) {
```
smcalilly, line 13: this was causing point types to blow up since they don't have the `setStyle` function

asset_dashboard/static/js/components/maps/SelectAssetsMap.js
```diff
@@ -189,14 +189,16 @@ function SelectAssetsMap(props) {
...
-          setSingleFeature(null),
+          setSingleFeature(null)
           
           // Reset the fillColor because the layer changed color 
           // whenever it was clicked on.
-          layer.setStyle({
-            fillColor: 'black',
-            fillOpacity: '0.2',
-          })
+          if (layer.setStyle) {
```
smcalilly, line 196: ditto on point types

fgregg (approved):

Issue #170, opened 2022-04-20
Filter CIP by section


smcalilly:

smcalilly:
When filtering by section, do we need to update the summary panels? 

For example, when all sections are showing up, it shows a total of all the section in the summary panels. When "civil engineering" is selected, do we need to show a summary of the costs for civil engineering?

Issue #167, opened 2022-04-12
Improve CIP summary information panels


smcalilly:
- [x] Add: Total estimated Cost for a given fiscal year (from Update Phase window)
- [x] Add: Total funded amount for a given fiscal year (from Phase Funding window)
- [x] Add: Total estimated cost by zone for a given fiscal year
- [x] Remove Projects
- [x] Remove Zones

Issue #166, opened 2022-04-12
Improve CIP table


smcalilly:
Add these columns:
- [x] Name
- [x] Description
- [x] Zone
- [x] Phase
- [x] Year
- [x] Quarter
- [x] Cost (This comes from Update Phase window)
- [ ] Secured Funding (This comes from Phase Funding window)
- [x] Score
- [x] Funding Source (? this was my idea when they were talking about grants and secured funding)
- [  ] make sure CSV export creates unique rows for each funding stream

smcalilly:
@fgregg Do you have any thoughts on how to represent these rows in a table? The listed columns are a combination of both a phase and a funding source, so these could be rows segmented by funding or rows combined by phase. We also have a requirement of exporting this data nicely into a CSV.

Some ideas:
1. Have phase as a row and then the funding sources be subrows, [kind of like this](https://www.google.com/search?q=table+with+rows+and+subrows&tbm=isch&ved=2ahUKEwjB4KOvq4_3AhWgQkIHHcIqAzwQ2-cCegQIABAA&oq=table+with+rows+and+subrows&gs_lcp=CgNpbWcQA1CbBFieDWDlD2gAcAB4AIABSogBuQWSAQIxMZgBAKABAaoBC2d3cy13aXotaW1nwAEB&sclient=img&ei=r91VYoHNJaCFieoPwtWM4AM&bih=676&biw=1423&client=firefox-b-1-d#imgrc=qOM41zg4A0qqRM).
    - [Our table library can do this](https://react-table.tanstack.com/docs/examples/sub-components) but it clicks to open the subrows
    - This might be able to export to CSV the way we want it (not sure right now but I bet we could figure it out)
2. List each phase and then combine the funding details into a funding column
    - In a funding column, include some sort of drop down or popup that you can click/hover over to show details about the funding
3. Treat each funding source as an individual row
    - This wouldn't work if a phase didn't have funding, unless we created a sensible default. (Though at the CIP planning stage could we assume there would always be some funding source?)
    - This way would export to csv better
    - It would look kind of like this:
       - project_1,phase_1,description,**funding_stream_1**,etc
       - project_1,phase_1,description,**funding_stream_2**,etc
       - project_1,phase_2,description,**no funding**,etc
       - project_1,phase_3,description,**funding_stream_21**,etc

fgregg:
We should export the data to csv as 3 (because it will be better for filtering and sorting)

For the website, i like 1.

[...]

PR #171, opened 2022-04-22
Improve the CIP + some minor changes

smcalilly:
## Overview

Closes:
- #166 
- #167
- #170

Includes:
- changes in the CIP UI
    - added columns to the tables
    - added subrows to the tables to show the phase funding
        - some minor code tweaks to accommodate this
    - calculations for the portfolio summary
- backend:
    - calculations for the zone distribution via a new signal + tests
    - refactored the local asset signal (added a delete signal and refactored the code) 
    - added data to send in the portfolio view
- minor bug fixes discovered along the way (noted/explained in this PR's comments)
- Added an `actual_cost` field to the phase model based on feedback

## Testing Instructions
Apparently review apps don't work with heroku and github anymore? So download this branch and follow the steps in the readme to get it running locally. 

For testing:
- Create a phase or two.
    - Add some GIS assets.
    - Add some funding.
- Test the CIP planner. Make sure all of the calculations work, particularly the zone distributions.
    - Add/remove phases
    - Test the funding dropdown subrows
    - Test the name and saving of a portfolio


asset_dashboard/management/commands/import_boundaries.py
```diff
@@ -37,10 +37,15 @@ def load_zones(self, geojson: dict, file_path: str):
...
         self.stdout.write(f'Importing {model} boundaries from {file_path}.')
+        print('model', model)
 
         for feature in geojson['features']:
             # house & senate is 'DISTRICT_INT' and commissioner is 'District_1'
-            district = feature['properties'].get('DISTRICT_INT', 'District_1')
+            district = feature['properties'].get('DISTRICT_INT')
+            
+            if not district:
+                district = feature['properties'].get('District_1')
```
smcalilly, line 47: dumb bug!

asset_dashboard/forms.py
```diff
@@ -106,6 +106,9 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        
+        for field_name, field in self.fields.items():
+            field.widget.attrs['required'] = True
```
smcalilly, line 111: the phase form was able to save without inputting data.

asset_dashboard/static/js/components/PortfolioTable.js
```diff
@@ -1,20 +1,17 @@
...
-      onClick: e => {
-        onRemoveFromPortfolio(original)
-      }
-    }
-  }
+  const [edit, setEdit] = useState(false)
+  
+  useEffect(() => {
+    setEdit(portfolio.id == null ? true : false)
+  }, [portfolio])
```
smcalilly, line 14: fixed this so the text input wasn't active when the page loads with an existing portfolio.

asset_dashboard/static/js/components/PortfolioTable.js
```diff
@@ -1,20 +1,17 @@
...
 import Button from 'react-bootstrap/Button'
 import Form from 'react-bootstrap/Form'
 import InputGroup from 'react-bootstrap/InputGroup'
 import projectColumns from './table_utils/projectColumns'
+import SubRow from './table_utils/SubRow'
 
 const PortfolioTable = ({ portfolio, onRemoveFromPortfolio, savePortfolioName, savePortfolio, createNewPortfolio }) => {
-  const [edit, setEdit] = useState(portfolio.name ? false : true)
-
-  const onRowClick = ({ original }) => {
```
smcalilly, line 11: needed to remove this so the subrow click could open. before this code change, clicking the entire row would add/remove the row from the portfolio. this made it impossible to click the subrow dropdown button.

asset_dashboard/static/js/components/PortfolioTable.js
```diff
@@ -28,11 +25,15 @@ const PortfolioTable = ({ portfolio, onRemoveFromPortfolio, savePortfolioName, s
...
-      <span>
-        <i className="fa fa-minus-square bg-"></i>
-      </span>
+      <button 
+        class='btn'
+        type='button'
+        onClick={() => onRemoveFromPortfolio(row)} 
+        aria-label='Remove project from portfolio'>
+          <i className="fa fa-minus-square fa-lg"></i>
+      </button>
```
smcalilly, line 36: this replaced the `onRowClick` function that i removed above.

asset_dashboard/static/js/components/ProjectsTable.js
```diff
@@ -1,21 +1,17 @@
 import React from 'react'
 import ReactTable from './BaseTable'
 import projectColumns from './table_utils/projectColumns'
+import SubRow from './table_utils/SubRow'
 
 const ProjectsTable = ({ allProjects, onAddToPortfolio, searchInput }) => {
-  const onRowClick = ({ original }) => {
```
smcalilly, line 6: same thing as [this comment](https://github.com/fpdcc/ccfp-asset-dashboard/pull/171/files#r856438891).

asset_dashboard/views.py
```diff
@@ -132,7 +143,7 @@ def form_valid(self, form):
             project = Project.objects.create(**project_data)
             ProjectScore.objects.get_or_create(project=project)
 
-            messages.success(self.request, 'Project successfully created!')
+            messages.success(self.request, 'Project successfully created.')
```
smcalilly, line 146: remove the exclamation points because it doesn't really fit the tone.

asset_dashboard/views.py
```diff
@@ -237,8 +248,8 @@ def form_valid(self, form):
         if form.is_valid():
             phase = form.save()
 
-            messages.success(self.request, 'Phase successfully edited!')
-            return HttpResponseRedirect(reverse('project-phases-list', kwargs={'pk': phase.project.pk}))
+            messages.success(self.request, 'Phase successfully updated.')
+            return HttpResponseRedirect(reverse('edit-phase', kwargs={'pk': phase.pk}))
```
smcalilly, line 252: found a bug.

asset_dashboard/static/js/PortfolioPlanner.js
```diff
@@ -298,10 +306,70 @@ class PortfolioPlanner extends React.Component {
   calculateTotals(portfolio) {
     return {
       budgetImpact: portfolio.reduce((total, project) => { return total + project.budget }, 0),
-      projectNames: portfolio.map(project => project.name),
-      projectZones: portfolio.map(project => project.zones.split(','))
+      totalEstimatedCostByYear: this.calculateEstimatedCostByKey(portfolio, 'year', 'total_estimated_cost'),
+      totalFundedAmountByYear: this.calculateFundedAmountByYear(portfolio),
+      totalEstimatedZoneCostByYear: this.calculateZoneCostByYear(portfolio)
```
smcalilly, line 311: these are calculated on the fly as projects are added/removed from the portfolio. it uses numbers calculated on the backend. would be nice to do it all on the backend but that would require a more significant change.

asset_dashboard/models.py
```diff
@@ -127,31 +129,80 @@ class Project(models.Model):
...
+            zone_distribution.zone_distribution_percentage = percentage
+            zone_distribution.save()
+
+    @classmethod
+    def calculate_zone_percentage(cls, distributions_by_zone, total_distribution_area):
+        percentages_by_zone = {}
+
+        for zone, distribution in distributions_by_zone.items():
+            quotient = distribution / total_distribution_area
+            percent = quotient * 100
```
fgregg, line 199: i would leave this out, since later you have to redived by 100 to get the right calculations. if you want to make the percent look good, then you should do that at the edge with a something like this: 

https://www.adamsmith.haus/python/answers/how-to-format-a-number-as-a-percentage-in-python

asset_dashboard/models.py
```diff
@@ -300,6 +370,76 @@ class LocalAsset(models.Model):
...
+                    geometrytype(geom) LIKE 'LINESTRING'
+                        OR geometrytype(geom) LIKE 'MULTILINESTRING'
+                   """]).aggregate(models.Union('geom'))['geom__union']
+
+        if assets:
+            # Return a buffered geometry so we can use the area of the buffer.
+            # The Geometries are in degrees, so use the 6th decimal place for
+            # creating the buffer. See https://gis.stackexchange.com/a/8674
+            # and https://docs.djangoproject.com/en/3.1/ref/contrib/gis/geos/#django.contrib.gis.geos.GEOSGeometry.buffer
+            return assets.buffer(.000005)
```
fgregg, line 428: let's make these buffer amounts settings that can be adjusted in the settings.py file.

asset_dashboard/static/js/PortfolioPlanner.js
```diff
@@ -332,25 +405,40 @@ class PortfolioPlanner extends React.Component {
...
+    })
+  }
 
   render() {
-    // This filters on every re-render, so that this.state.remainingProjects can be the source of truth.
-    // Could also chain filtering here for other things (like "department" from the wireframe).
     const filteredRows = this.state.remainingProjects && this.state.remainingProjects.filter(project => {
       return project.description.toLowerCase().includes(this.state.filterText.toLowerCase())
+    }).filter(project => {
+      if (this.state.selectedSection !== '') {
```
fgregg, line 427: why not 

```
if (this.state.selectedSection) {
```

asset_dashboard/static/js/PortfolioPlanner.js
```diff
@@ -332,25 +405,40 @@ class PortfolioPlanner extends React.Component {
...
+        return project.section.toLowerCase().includes(this.state.selectedSection.toLowerCase())
+      } else {
+        return project
+      }
+    })
+    
+    const filteredPortfolio = this.state.portfolio.projects && this.state.portfolio.projects.filter(project => {
+      return project
+    }).filter(project => {
+      if (this.state.selectedSection !== '') {
```
fgregg, line 437: this seems to be an invariant, maybe lift this up?

asset_dashboard/static/js/PortfolioPlanner.js
```diff
@@ -332,25 +405,40 @@ class PortfolioPlanner extends React.Component {
...
       return project.description.toLowerCase().includes(this.state.filterText.toLowerCase())
+    }).filter(project => {
+      if (this.state.selectedSection !== '') {
+        return project.section.toLowerCase().includes(this.state.selectedSection.toLowerCase())
+      } else {
+        return project
+      }
+    })
+    
+    const filteredPortfolio = this.state.portfolio.projects && this.state.portfolio.projects.filter(project => {
```
fgregg, line 434: what about optional chaining here:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

asset_dashboard/static/js/PortfolioPlanner.js
```diff
@@ -332,25 +405,40 @@ class PortfolioPlanner extends React.Component {
...
+      return project
+    }).filter(project => {
+      if (this.state.selectedSection !== '') {
+        return project.section.toLowerCase().includes(this.state.selectedSection.toLowerCase())
```
fgregg, line 438: you've used this construct a few times.

what about creating a little helper function

```
function caseInsensitiveIncludes(source, target) {
     return source.toLowerCase().includes(target.toLowerCase())
```

fgregg (changes requested):

asset_dashboard/static/js/PortfolioPlanner.js
smcalilly: optional chaining would [return undefined if we used it on an array method](https://stackoverflow.com/a/60165273). i'm afraid that might blow up the table since the table needs an iterable, but i'll try. it's possible that we don't need any truthy check whatsoever, since this code is an artifact from the CIP prototype.

fgregg: i see. wouldn't your code return an undefined right now if projects was not defined?

asset_dashboard/static/js/PortfolioPlanner.js
fgregg: go for it.

asset_dashboard/models.py
smcalilly: oh cool! so just let it hangout in the database as .75 instead of 75%? makes better sense.

asset_dashboard/static/js/PortfolioPlanner.js
smcalilly: good point. i'll rework this.

asset_dashboard/static/js/PortfolioPlanner.js
smcalilly: see comment here: https://github.com/fpdcc/ccfp-asset-dashboard/pull/171#discussion_r858945389

asset_dashboard/static/js/PortfolioPlanner.js
smcalilly: done so here: https://github.com/fpdcc/ccfp-asset-dashboard/commit/aa9691fa5b92f138284e8198b24feaba889d84c8

fgregg (approved):

Issue #181, opened 2022-05-12
add funding secured to CIP portfolio export


glw:
Add funding "secured" column to CIP portfolio export.

Issue #180, opened 2022-05-12
fix phase form validation and conditionals


smcalilly:
have "actual cost" displayed even if it’s not “complete”. turn of form validation for that field

fix year input form — it doesn't turn off validation if the status is "complete" and therefore there is no way to save a phase as "completed" within the current year or retroactively

Issue #175, opened 2022-05-03
give notification when nothing is found in assets


glw:
Give some kind of feedback if searching an asset and nothing is found or returned.

![image](https://user-images.githubusercontent.com/556205/166555011-118b6d1f-766c-485a-9e32-42dfd874ed1f.png)

Issue #174, opened 2022-05-03
unable to remove an asset where there is only one left


glw:
You are able to delete assets until you reach the final asset at that point it wont let you remove the last one.

![image](https://user-images.githubusercontent.com/556205/166504170-0982b966-6ffb-4baa-bfa3-05926ddc9d35.png)

PR #184, opened 2022-05-13
Fix various bugs + improvements from user feedback

smcalilly:
## Overview
Closes:

- #183 
- #174
- #181 
- #175
- #180

## Testing Instructions
- follow readme instructions to load gis data if you don't already have it
- make sure all the money values have no decimal (# 183)
    - this'll be on the project detail page, update phase page, create/update funding stream, cip planner
 - add some assets and make sure you can delete all of them (# 147)
 - export a csv from the cip planner and make sure the csv includes a funding_secured column 
 - duplicate the gis search from (#175) and make sure a message appears
 - for #180
     - save a phase as new and don't input a number into the actual cost field. it should save since we're not validating that field but we're showing it anyway.
     - set a phase as "complete" and try to input this year or older (because some js removes the minimum input validation)


asset_dashboard/models.py
```diff
@@ -195,6 +195,9 @@ def save_zone_distribution(sender, instance, **kwargs):
     def calculate_zone_proportion(cls, distributions_by_zone, total_distribution_area):
         proportions_by_zone = {}
 
+        if total_distribution_area == 0.0:
+            return distributions_by_zone
```
smcalilly, line 199: #174 was due to a backend failure. the code would delete the asset and then try to divide by 0, aka `ZeroDivisionError: division by zero`

asset_dashboard/static/js/components/maps/SelectAssetsMap.js
```diff
@@ -132,6 +132,11 @@ function SelectAssetsMap(props) {
     }).then((response) => response.json())
     .then((data) => {
       setIsLoading(false)
+      
+      if (data.features.length == 0) {
+        setAjaxMessage({text: 'No assets found with search query.', tag: 'info'})
+      }
```
smcalilly, line 138: for #175

asset_dashboard/views.py
```diff
@@ -179,6 +179,13 @@ def get_context_data(self, **kwargs):
                     'assets': LocalAssetReadSerializer(assets, many=True).data
                 }
 
+            assets = LocalAsset.objects.filter(phase__project=self.object)
+
+            if assets.exists():
+                context['props'] = {
+                    'assets': LocalAssetReadSerializer(assets, many=True).data
+                }
```
smcalilly, line 187: dunno how this got pulled in from PR #149 because i rebased

fgregg (approved):

Issue #202, opened 2022-06-14
investigate weighted score


smcalilly:
weighted score average should be between 0 and 5.

smcalilly:
@fgregg here's the relevant code: https://github.com/fpdcc/ccfp-asset-dashboard/blob/master/asset_dashboard/models.py#L336-L355

it looks like it's calculating the total score as expected. here is a stripped down version of that code in a python shell, using the same values of the project where we saw this bug in the demo:
```python
>>> scores = [2, 2, 3, 4, 2, 0]
>>> weights = [0.5, 0.5, 0.3, 0.2, 0.3, 0.3]
>>> total = 0
>>> for score, weight in zip(scores, weights):
...     total += score * weight
...
>>> print(total)
4.3
```

that `4.3` equals [what is on the staging site](https://ccfp-asset-dashboard-staging.herokuapp.com/projects/45/). it would equal > 5 if you tweak the values.

what do we need to change in the calculation so the total is <= 5?

fgregg:
it needs to be

```
for score, weight in zip(scores, weights):
     total += score * weight

total /= sum(weights)

PR #203, opened 2022-06-14
Fix score weight total

smcalilly:
## Overview

- Closes: #202

### Notes

To fix this, we're dividing the total by the sum of the score weights.

## Testing Instructions
- visit review app: https://ccfp-asset-d-fix-score--lhc8v8.herokuapp.com/
- see some project detail pages
- confirm the weight score is less than or equal to 5

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -38,7 +38,7 @@ def test_project_score_total_method(project, score_weights):
         score_field_value = getattr(project_score_instance, score_field_name)
         total_score += score_field_value * weight_field_value
 
-    assert project_score_instance.total_score == total_score / weights_sum
+    assert int(project_score_instance.total_score) == int(total_score / weights_sum)
 
     # test that an error will raise if there is no score weight
     with pytest.raises(ValueError):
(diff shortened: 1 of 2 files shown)


tests/test_models.py
```diff
@@ -30,12 +30,15 @@ def test_project_score_total_method(project, score_weights):
...
+    weights_sum = 0
     for index, score_field_name in enumerate(updated_scores):
         weight_field_value = getattr(score_weights, score_field_name)
+        weights_sum += weight_field_value
+        
         score_field_value = getattr(project_score_instance, score_field_name)
         total_score += score_field_value * weight_field_value
 
-    assert project_score_instance.total_score == total_score
+    assert int(project_score_instance.total_score) == int(total_score / weights_sum)
```
smcalilly, line 41: this failed when trying to assert 3.8437500000000004 == 3.84375, so i cast to an int

in hindsight, this test seems kind of redundant and pointless since it's replicating the code that it's testing. let me know if you can think of a better way to test this?

fgregg: it's a fine type of test since the implementation might change.

checkout out this pytest method for dealing with just such issues: https://docs.pytest.org/en/latest/reference/reference.html?highlight=approx#pytest.approx

fgregg (commented):
use the approx method in the test, and we'll be good to go!

smcalilly:
done here: eee5417

PR #211, opened 2022-07-19
set score_field_value as 0 instead of returning 0 as the total score

smcalilly:
## Overview

- Closes #208  

### Notes
This change:
- defaults all the project score fields when a project is created
- adds a signal to set `geographic_distance_score` and `social_equity_score` to 5 when the project is countywide

## Testing Instructions
- visit review app: https://ccfp-asset-d-patch-scor-fafe7c.herokuapp.com
- create a project
- change the scores
- confirm that the score changes even if there are no gis assets connected to the project

to test the countywide change:
- create a phase
- set the project as countywide
- confirm the score updates in the project detail view

PR #216, opened 2022-08-03
Fix funding delete bug

### LINKED ISSUE #212: production - Phase funding can not be deleted
After adding a phase funding item it can not be deleted. Takes you to "page not found".
cip_prod_delete_phase_funding

smcalilly:
## Overview

- Closes #212
- Fixes issue found after merging PR #211 

### Demo
phase-funding-fix

### Notes
The delete button used the pk for the phase and not the funding stream, so it led to a 404 not found.

I realized that the fix from PR #211 won't affect current scores. This fix works so it's backwards compatible, but should we maybe write a script to update the database and remove all `None` values from all ProjectScore rows?

## Testing Instructions
- visit review app tk
- go to a phase
- delete funding
- make sure it redirects to a confirmation screen, and then back to the correct phase


fgregg (approved):
looks fine. i might just shell in and manually fix?

PR #231, opened 2023-05-08
Allow for the bid quarter to not be required

### LINKED ISSUE #204: allow bid qtr null value
Allow users the option to leave bid quarter blank.

xmedr:
## Overview

This allows for users to not have to pick anything for the bid quarter when creating/editing a phase.

The field in the model already allowed for this to be blank/null, so this just makes it optional in the form, and adjusts areas where the name is dependent on the bid quarter. In those cases, the quarter is omitted all together, so the name would be `preliminary_engineering - 2025 - new` as opposed to `preliminary_engineering - null - 2025 - new`. I can absolutely change this if we want to!

Closes #204
### Demo

image
---
image

## Testing Instructions

* Create a phase and save it without having chosen a bid quarter
* Confirm that the phase has been created and that any display names do not include null or none

### DIFF
--- a/asset_dashboard/forms.py
+++ b/asset_dashboard/forms.py
@@ -100,3 +100,6 @@ class PhaseForm(StyledFormMixin, ModelForm):
 
         for field_name, field in self.fields.items():
             field.widget.attrs['required'] = True
+
+            if field_name == 'estimated_bid_quarter':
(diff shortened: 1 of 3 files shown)


smcalilly (approved):
nice!

PR #235, opened 2023-05-10
Fix phase / asset GIS bug

### LINKED ISSUE #223: when saving assets, existing districts were overwritten
On staging, I saved assets to a phase and the existing districts were overwritten. 

I'm not sure if this is a staging data issue or what.

- [ ] recreate locally
- [ ] debug
- [ ] fix

### LINKED ISSUE #224: 500 when adding assets to the phase
On staging, I sometimes get a 500 error when adding assets to the phase.

smcalilly:
## Overview

- closes #224 
- closes #223 

includes:
- gis bug fix (#224)
- districts overwritten fix (#223)
- refactoring of the signals (see notes)
- tests (~half of the diff is a geojson test fixture)
- upgraded django to the latest patch version (see latest security release: https://www.djangoproject.com/weblog/2023/may/03/security-releases/)
- upgrade python (3.8 -> 3.9)

#### gis bug fix
The comments in that issue include a lot of background about that bug, but TLDR: the intersects query was causing a segmentation fault on posgres when trying to find the intersection of political districts and a phase's assets. This happened sometimes, but not always. To fix, I'm filtering first with the `dwithin` function, then doing the intersection. The `dwithin` function trims down the area where we want to find the intersection. I've added a comment to this PR so you can see the dwithin change, since this diff is bigger with the refactor.

Update: I've upgraded postgis in the docker container so this is fixed on local development. We're migrating to CCFP's servers, so we'll hold off on fixing this on heroku (see this comment: https://github.com/fpdcc/ccfp-asset-dashboard/issues/224#issuecomment-1549877563)

#### signal refactor
Also includes a refactor of the django signal and GIS processing code. This fixes #223.

I've refactored all of the GIS logic into a class. I also found that a lot of the work was duplicated across signals, so I've consolidated all of the asset-related signals into one single signal so there's not as much redundant work ([see this comment in the bug issue about the inefficient, redundant code](https://github.com/fpdcc/ccfp-asset-dashboard/issues/224#issuecomment-1538948373)). This also makes all of the GIS processing logic easier to debug, follow, etc, since it's all in one place rather than split across signals.

## Testing Instructions
- visit this project on the review app: https://ccfp-asset-d-patch-sign-5yfa9e.herokuapp.com/projects/2/
  - login with temp credentials: username=`admin` && password=`password`
- in the project detail view linked above, scroll down and click "edit assets"
- search for an asset in cook county. bonus points if you find one that is geographically far away from the existing assets of that phase.
- add those assets to the phase. this action should show a success message.
- navigate back to the project overview (click the "overview" button). make sure the scores updated, specifically the political districts, geographic distance score, and social equity score


asset_dashboard/models.py
```diff
@@ -129,91 +130,39 @@ class Project(models.Model):
...
+    def update_project_districts(self, phase_geoms):
         district_models = [
             ("commissioner_districts", CommissionerDistrict),
             ("senate_districts", SenateDistrict),
             ("house_districts", HouseDistrict),
         ]
 
         for attribute, model in district_models:
-            districts = model.objects.filter(boundary__intersects=phase_geoms)
+            districts = model.objects.filter(boundary__dwithin=(phase_geoms, D(m=1))).filter(boundary__intersects=phase_geoms)
```
smcalilly, line 151: this is the dwithin bug fix

tests/test_views.py
smcalilly, line 1: had to change these when upgrading django

xmedr (commented):
@smcalilly Heads up, I found an asset with an id of 86, and upon adding it, it did say it was saving the asset but I didn't get a success message. Then afterwards, I clicked the overview button and got a page that just read "Internal Server Error".

However, refreshing this page did bring me back to the project detail page, where I could see that new asset under the list of existing ones. But clicking edit assets from here gives me another server error. I could consistently reproduce this bug by deleting and readding that asset and another one I tried.

smcalilly:
@xmedr woof. thanks for flagging. it's the same memory error that i thought i had fixed. :(

asset_dashboard/models.py
fgregg: i find it is suprising that this is using less memory than the boundary_intersects.

can you post the districts.query generated SQL for each of these?

smcalilly:
@fgregg I upgraded the postgis version for local development and that upgrade GEOS. The dwithin is a workaround but it doesn't address the underlying GEOS issue. Should we merge this code as is or remove dwithin?

PR #242, opened 2023-06-22
Fix trails <-> trails info relationship

### LINKED ISSUE #241: Trails gis query throwing an error
From an email:

> One issue I already encountered was I couldn’t select trail data when I was creating new projects. I was able to see the selected segments in existing projects though. So please take a look at that.

I was able to recreate this on production.

smcalilly:
## Overview

The schema changed to where the trails_info_id is a column in the trails table, so we need to use that new relationship to access the trails geometry data.

- closes #241 

other changes include:
- added `trail_name` and `regional_trail_name` to the `Trail` query
- added `pointsofinterest_id__isnull` to the `PoiInfo` search query (found that bug when testing this change)
- catch `ParkingLots.DoesNotExist` exception (this was an edge case I found when testing this PR's change — one of the parking_lot_ids is 0 and that record doesn't exist...)
- upgraded some dependencies

### Notes
We need to make sure other tables haven't changed. Here's the latest diagram from @jim-z 
quercus-6-22-2023

I opened issue #243 to do this.

## Testing Instructions
- visit the map view and search for some trails: https://ccfp-asset-d-patch-trai-msqvfj.herokuapp.com/projects/phases/edit/2/assets/
  - username=`testuser` & password=`badpassword`


fgregg (approved):
nice work.

PR #244, opened 2023-07-10
Feature/phase and funding form

### LINKED ISSUE #156: Phase and funding forms on one page

xmedr:
## Overview

This allows for the manipulation of multiple funding streams from a phase edit page. Features include:
- adding multiple new fundings at once
- editing multiple existing fundings at once
- warning the user if new fundings have not been saved when navigating away
- front and back end validation of values

Closes #156 

### Demo
image

### Notes

The funding table scrolls horizontally on small screen sizes. This is the best display I could come up with, but if there is a more appealing way of fitting the table please lmk!

## Testing Instructions
* Create a project and add a phase
  * Confirm that the Update Funding and Add a Source buttons don't render yet while on the creation page 
* After creating a phase, use the Add a source button to add a few rows
* Use the Remove button to get rid of some of the rows
* Test the forms with valid and invalid inputs and hit the Update button
  * Confirm that error messages show appropriately for invalid inputs
  * Confirm that error messages disappear once fixed
* The page should reload after hitting update, when all inputs are valid
  * Confirm that your inputs are saved (indicated by a Delete button next to the fundings)
* Add another a new row, fill in some values, and refresh the page
  * Confirm that a dialog box warns about unsaved changes on the page


xmedr:
The way that django is rendering these forms, there are multiple input elements with the same id like `id_year`, `id_budget_0`, etc. I tried instantiating the existing funding forms with different ids based off of their primary keys, but then couldn't figure out a solid way of selecting them to be invisible in css. So everything works regardless, but as of now I haven't found an elegant way to resolve that.

Also as of now, having a budget and actual cost of 0 is an invalid input. If zero can be a valid value for this field then I can definitely change this!

asset_dashboard/models.py
```diff
@@ -420,6 +420,10 @@ class FundingStream(models.Model):
         null=True,
     )
 
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        return super().save(*args, **kwargs)
```
smcalilly, line 425: were you not able to get the validation to work without this? i'm wondering if `full_clean` is already called whenever the form validation happens?

asset_dashboard/serializers.py
```diff
@@ -280,3 +280,71 @@ def save(self):
...
+    def save(self):
+        data = {i: self.validated_data[i] for i in self.validated_data if i != 'phase'}
+        phase = Phase.objects.get(id=self.validated_data["phase"].id)
+        funding_stream, created = FundingStream.objects.update_or_create(
+            id=self.validated_data["id"],
+            defaults=data
+        )
+        phase.funding_streams.add(funding_stream)
+
+        return data
```
smcalilly, line 350: nice use of dict comprehension and update_or_create1

asset_dashboard/templates/asset_dashboard/partials/forms/add_edit_phase_form.html
```diff
@@ -48,47 +48,72 @@ <h3>{% if phase %}Update Phase{% else %}Create Phase{% endif %}</h3>
     <div class="d-flex justify-content-between m-3">
       <h3>Phase Funding</h3>
       {% if phase %}
-        <a href="{% url 'create-funding' pk=phase.id %}" class="text-info lead">Add funding ></a>
+        <a href='#/' class='btn btn-info' onclick='submitFundingForms()'>Update Funding</a>
```
smcalilly, line 51: a couple of things here:

1. this should be a `<button>` instead of an `<a>` tag. that way you don't have to add that href and you can just attach the click event to the button.
2. instead of attaching the `submitFundingForms` function to the onclick, what about adding an id to this element and in your javascript you can grab that id to add an event listener for click events? [this shows an example of how to do that](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#examples). it has the added benefit of keeping all of the javascript contained within the scripts.

asset_dashboard/templates/asset_dashboard/partials/forms/add_edit_phase_form.html
```diff
@@ -129,6 +154,239 @@ <h3>Phase Assets</h3>
...
+            }).then(data => {
+              if (!backendValid) {
+                showBackendErrors(data, ...errors)
+              }
+            })
+          )
+        }
+
+        // Wait until all fetches run, then either show serverError or reload page
+        Promise.all(fetches).then(function() {
```
smcalilly, line 312: this is a clever way to handle this — i like!

asset_dashboard/templates/asset_dashboard/partials/forms/add_edit_phase_form.html
```diff
@@ -129,6 +154,239 @@ <h3>Phase Assets</h3>
...
+          // Ask user if they want to leave page, if there are unsaved funding inputs
+          e.returnValue = ""
+        }
+          
+      }
+    }
+
+    function showBackendErrors(data, yearError, budgetError, costError, sourceError, securedError) {
+      let keys = Object.keys(data)
+      keys.forEach(key => {
```
smcalilly, line 355: you can just do `Object.keys(data).forEach` without creating the keys variable, but that's a stylistic thing that is up to you.

asset_dashboard/templates/asset_dashboard/partials/forms/add_edit_phase_form.html
```diff
@@ -129,6 +154,239 @@ <h3>Phase Assets</h3>
...
+          hideError(costError)
+        }
+      }
+
+      // Back end validation and submission of forms
+      if (valid) {
+        let backendValid = true
+        let fetches = []
+
+        for (i = 0; i < fundings.length; i++) {
```
smcalilly, line 261: did you try a .forEach or .map here? my hunch is that the async nature of fetch might not work well with those array functions (i've had weird problems trying to do async stuff within forEach).

just curious if you tried, no change necessary.

smcalilly (changes requested):
@xmedr this is great! good work. i left a few comments inline, nothing too significant.

the biggest piece of feedback i have is UX related. the navigation warning comes up whenever i try to navigate away from the page in any way. 

for example, i get that message if i go to create a new phase and try to "save phase". here's a gif that shows that:
phase-demo

in the case of creating a new phase, that warning shouldn't come up. i can see the use of that for some other interactions, however. this might warrant a bigger discussion with everybody, but it feels like that should only show up whenever the funding forms change without being saved. do you have thoughts about this?

you might be able to add an event listener to the phase funding forms that listen to anytime the forms change, in which case you could add the event listener for the navigation warning.

asset_dashboard/serializers.py
```diff
@@ -280,3 +280,71 @@ def save(self):
...
+
+    def validate_source_type(self, value):
+        choices = [
+            "capital_improvement_fund",
+            "bonds",
+            "construction_development",
+            "grants_fees_other",
+            "rollover",
+        ]
+        if value not in choices:
```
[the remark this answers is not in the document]
xmedr: ohh cool, i see what you mean. that code worked perfectly, ty!

asset_dashboard/models.py
xmedr: oh you're right! i think i was having issues with validation not running when i was applying validators directly to the models, and this might have helped?

but regardless, i'm doing it through the serializer now and it works without this save override, so i've removed this

asset_dashboard/templates/asset_dashboard/partials/forms/add_edit_phase_form.html
xmedr: ohh nice, both of these make a lot of sense.

asset_dashboard/templates/asset_dashboard/partials/forms/add_edit_phase_form.html
xmedr: it looks like since i'm grabbing this with `getElementsByClassName` it's returning an htmlcollection, and i remember trying `foreach` with an htmlcollection before (not sure if here or elsewhere) and it not working.

turns out those don't have `foreach` or `map` methods, [as opposed to a nodelist](https://www.w3schools.com/js/js_htmldom_nodelist.asp#:~:text=The%20querySelectorAll()%20method%20returns,property%20returns%20a%20live%20NodeList.) which i could've gotten if i used `querySelectorAll` instead. i never knew this was the reason, TIL

asset_dashboard/templates/asset_dashboard/partials/forms/add_edit_phase_form.html
xmedr: thanks! i can see myself definitely coming back to this way of waiting for multiple fetches in the future haha

asset_dashboard/templates/asset_dashboard/partials/forms/add_edit_phase_form.html
```diff
@@ -129,6 +154,244 @@ <h3>Phase Assets</h3>
...
+
+      for(i = 0; i < newFormsContainers.length; i++ ){
+        let yearInput = newFormsContainers[i].querySelector('#id_year').value
+        let budgetInput = newFormsContainers[i].querySelector('#id_budget_0').value
+        let costInput = newFormsContainers[i].querySelector('#id_actual_cost_0').value
+        
+        if(yearInput || budgetInput > 0 || costInput > 0) {
+          // Ask user if they want to leave page, if there are unsaved funding inputs
+          e.preventDefault();
+          e.returnValue = ""
```
xmedr, line 354: @smcalilly Alright i've got amongst other things, a change in there that hopefully resolves that unsaved-changes-warning weirdness.

I previously had this `e.preventDefault` at the start of this `redirectDialog` function, but it looks like putting it a bit later in this `if` still prevents unloading and gives us what we're looking for

smcalilly (changes requested):
@xmedr your prevent default fix worked!

one tiny change and then this can come in. we should allow 0 values for both the funding amount and actual cost fields.

xmedr:
@smcalilly gotcha! I also realized that the model allowed negative dollar amounts, so i changed the serializer to disallow that. is that alright?

smcalilly:
@xmedr yep! good catch.

PR #247, opened 2023-07-12
Allow for a project score of 0

### LINKED ISSUE #246: Fix scoring field validation project form
From @jim-z email:

> One user was noticing that they were able to add a project but the information they were adding wasn’t being saved.  I had them walk me through what they had done and it appears that they were hitting “save project” and navigating away from the page without noticing the forced constraints in the scoring section:

image003

> Can we disable the scoring requirement for now? We still haven’t determined how to carry out the scoring component internally so it isn’t going to be utilized this year.

This is happening because the score fields equal 0 when the form is submitted, which is invalid based on [the validators in the `ScoreField`](https://github.com/fpdcc/ccfp-asset-dashboard/blob/master/asset_dashboard/models.py#L305). We should fix the validation to allow 0.

We'll know this is fixed if we can edit a project where all of the scores equal 0, and we don't get the errors in the above screenshot.

xmedr:
## Overview

Quick fix to allow users to save a project with a score of 0

Closes #246 

## Testing Instructions
* Create/edit a project
* On the project detail page, set all scores to zero and save
* Confirm that the project page reloads, with the success message at the top


smcalilly (approved):
nice n easy. thanks!

Issue #252, opened 2023-07-20
Users unable to delete assets


hallj89:
https://ccfp-asset-dashboard-prod.herokuapp.com/projects/76/

Duplicate assets added accidentally and can't be removed. Delete button doesn't respond, or after refresh gives internal server error then successful delete message while not actually removing anything.

smcalilly:
from the logs:

```
2023-08-02T15:59:26.537819+00:00 app[web.1]: self.project.update_project_districts(self.phase_geometries)
2023-08-02T15:59:26.537819+00:00 app[web.1]: File "/app/asset_dashboard/models.py", line 156, in update_project_districts
2023-08-02T15:59:26.537819+00:00 app[web.1]: for district in districts:
2023-08-02T15:59:26.537820+00:00 app[web.1]: File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 280, in __iter__
2023-08-02T15:59:26.537820+00:00 app[web.1]: self._fetch_all()
2023-08-02T15:59:26.537820+00:00 app[web.1]: File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 1324, in _fetch_all
2023-08-02T15:59:26.537820+00:00 app[web.1]: self._result_cache = list(self._iterable_class(self))
2023-08-02T15:59:26.537821+00:00 app[web.1]: File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 51, in __iter__
2023-08-02T15:59:26.537821+00:00 app[web.1]: results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)
2023-08-02T15:59:26.537821+00:00 app[web.1]: File "/usr/local/lib/python3.8/site-packages/django/db/models/sql/compiler.py", line 1175, in execute_sql
2023-08-02T15:59:26.537821+00:00 app[web.1]: cursor.execute(sql, params)
2023-08-02T15:59:26.537823+00:00 app[web.1]: File "/usr/local/lib/python3.8/site-packages/sentry_sdk/integrations/django/__init__.py", line 499, in execute
2023-08-02T15:59:26.537823+00:00 app[web.1]: return real_execute(self, sql, params)
2023-08-02T15:59:26.537823+00:00 app[web.1]: File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 66, in execute
[...]

PR #267, opened 2023-08-09
Fix missing geo calculations

### LINKED ISSUE #265: Geography section not populating
Zone and district fields are blank for every project.

image

smcalilly:
## Overview
- closes #265 

Includes a Heroku review app config change to support a different database connection (to be addressed in another PR, but related to #252)

### Notes

This happened due to two things:
1. for the `Zone` model, the import was skipping the zones, so I changed it to an `update_or_create`
2. for districts, moving self.save() outside of the loop fixed this...dunno why

## Testing Instructions
- visit the review app: https://ccfp-asset-d-feature-di-1nlzy2.herokuapp.com/
- navigate to a project with some assets ([here's one](https://ccfp-asset-d-feature-di-1nlzy2.herokuapp.com/projects/428/)). see how the projects don't include the districts/zones
- add/remove an asset. visit the project overview page and see how the districts/zones get updated


asset_dashboard/models.py
```diff
@@ -155,7 +155,8 @@ def update_project_districts(self, phase_geoms):
 
             for district in districts:
                 project_district.add(district)
-                self.save()
+
+            self.save()
```
smcalilly, line 159: ouch! weird?

app.json
```diff
@@ -20,7 +23,7 @@
   },
   "environments": {
     "review": {
-      "addons": ["heroku-postgresql:hobby-basic"]
+      "addons": []
```
smcalilly, line 26: need to remove this to use an RDS database for staging/review apps.

fgregg (approved):

PR #269, opened 2023-08-15
Prevent preexisting assets from saving again

### LINKED ISSUE #266: Prevent assets from being duplicated
## Bug description

When selecting assets with the map draw tool, a user can re-select an asset that they've already selected. This duplicates the asset so that the asset shows up one or more times for a phase. 

Exposed in #252.

## Fix
If a user has already selected an asset, then some logic should prevent the same asset from being submitted to the API. This should happen in the javascript code that prepares the assets to be submitted to the API. The `SelectAssetsMap.js` component submits the data to the API, and [the `GeometrySelector` component](https://github.com/fpdcc/ccfp-asset-dashboard/blob/master/asset_dashboard/static/js/components/map_utils/GeometrySelector.js) parses out the assets from the selected region. You'll need to add some helper code to the `GeometrySelector` that compares the selected assets with the phase's existing assets, and only include the selected asset in the payload that's submitted to the API.

## Completed
This issue will be completed when a user can select and save the same asset in the map multiple times without duplicating the asset in the database.

xmedr:
## Overview
This branch fixes an asset duplication bug, so if a user:

 - selects only preexisting assets with the draw tool and saves, they will be presented with an error message at the top of the page and those assets will not be saved again.
 - selects a mix of preexisting and new assets, the preexisting ones will be excluded from the fetch command, allowing only the new assets to be saved.

Closes #266 

### Demo

image

## Testing Instructions
* Navigate to a project's phase and edit it's assets
* If needed, add a one or two assets
* Use the draw tool to select your previously added asset(s) along with some new ones
  * Confirm that only the new assets get saved, and the old ones are not duplicated
* Use the draw tool or the selector on the left to select only your previously added asset(s)
  * Confirm that nothing gets saved, the warning is displayed, and the page does not automatically reload

### DIFF
--- a/asset_dashboard/static/js/components/maps/SelectAssetsMap.js
+++ b/asset_dashboard/static/js/components/maps/SelectAssetsMap.js
@@ -70,15 +70,33 @@ function SelectAssetsMap(props) {
   }
 
   function saveGeometries() {
-    const data = geomsToSave['features'].map(feature => {
-      return {
-        'asset_id': feature['properties']['identifier'],
-        'asset_type': searchAssetType,
-        'asset_name': feature['properties']['name'],
-        'geom': feature['geometry'],
-        'phase': phaseId
+    let existingIDs = []
+    existingGeoms.features.forEach(geom => {
+      existingIDs.push(geom.properties.asset_id)
+    })
+
+    let data = geomsToSave['features'].map(feature => {
+      // Only include feature if it doesn't already exist
+      if (!existingIDs.includes(feature.properties.identifier.toString())) {
+        return {
+          'asset_id': feature['properties']['identifier'],
+          'asset_type': searchAssetType,
+          'asset_name': feature['properties']['name'],
+          'geom': feature['geometry'],
+          'phase': phaseId
+        }
       }
-    }) 
+    })
+
+    // map() adds 'undefined' if the conditional fails, so filter those out
+    data = data.filter((item) => {
+      return item !== undefined
+    })
+
+    if (data.length == 0) {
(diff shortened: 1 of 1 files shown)


asset_dashboard/static/js/components/maps/SelectAssetsMap.js
```diff
@@ -70,15 +70,33 @@ function SelectAssetsMap(props) {
...
+          'phase': phaseId
+        }
       }
-    }) 
+    })
+
+    // map() adds 'undefined' if the conditional fails, so filter those out
+    data = data.filter((item) => {
+      return item !== undefined
+    })
```
smcalilly, line 94: nice filter. you can chain it to the end of your map instead of resetting the `data` value. something like
```
    let data = geomsToSave['features'].map(feature => {
      // Only include feature if it doesn't already exist
      if (!existingIDs.includes(feature.properties.identifier.toString())) {
        return {
          'asset_id': feature['properties']['identifier'],
          'asset_type': searchAssetType,
          'asset_name': feature['properties']['name'],
          'geom': feature['geometry'],
          'phase': phaseId
        }
      }
    }) 
    }).filter((item) => {
      return item !== undefined
    })
```

smcalilly (commented):
looks good! feel free to make the suggested changes or not, that's up to you.

asset_dashboard/static/js/components/maps/SelectAssetsMap.js
xmedr: ah cool, this helps clean things up, ty!

Issue #257, opened 2023-07-25
Create additional filters for selecting projects within CIP Planner


jim-z:
Need the ability to sort projects by phase year.  Currently can only sort by "year" but annual CIP is primarily based on the more specific phased funding years/amounts.
![image](https://github.com/fpdcc/ccfp-asset-dashboard/assets/11774880/84ee4bb0-3c48-4012-841f-7d528d9edcde)

xmedr:
Short term: We'll be creating a data pool to help more easily create a 5-year plan by Thursday

Long term: We'll implement a way to sort phases by the year of their fundings, not just the year of the phase itself

smcalilly:
@jim-z just to clarify, the year sort is based on the phase year, and y'all need to be able to sort based on the phase funding year?

in that case, would it be okay if we changed the UI for this? specifically, from your above screenshot, the relationship between the project's phase and the sub-table with the phase funding. in order to sort, i think we'll need to flatten the rows so that each phase funding source creates a unique row instead of containing that data in a sub-table.

the table would roughly look like this:
|phase|phase_year|funding_source|funding_amount|funding_year|project_description|
|------|------------|---------------|-----------------|-------------|--------------------|
|phase_1|2023|grants|$200k|2023|Construction following feasibility study from Engineering Contract|
|phase_1|2023|capital improvements|$70k|2024|Construction following feasibility study from Engineering Contract|

note the duplication of data in the columns `phase`, `phase_year`, `project_description`. this change would create duplicated data in the CIP tables, including the other phase and project fields that i omitted for brevity. the unique data in each row would be the phase funding, and all of the phase and project fields would be duplicated across each row. we could figure out a way to truncate the longer text fields so it's a better experience visually.

this change would be the best way to structure this data so we can filter on it in the table, and structuring the data like that would make the data better work with it after y'all export it.

i'm open to other ways of addressing this problem, but as it's currently designed it's likely we'll need to hack it in order to get the sorting to work on the phase funding fields.

smcalilly:
@jim-z or, would a dedicated filter be a better solution? rather than using the table sort, we could create a filter similar to the "filter by section". if we did this then we might need to adjust the UI so those filters are placed in a way to that connects them to the table in a more obvious way.

jim-z:
[...]

PR #270, opened 2023-08-16
Flatten CIP tables based on phase funding streams + add more data to CIP export

### LINKED ISSUE #219: CIP portfolio asset summary
Create a way to export an asset summary by portfolio. Whether that be an image of a map and/or adding the asset ID's to the CIP export. This way users can get a visual location representation of their projects in a particular portfolio.

smcalilly:
## Overview

- related to #257, but doesn't close it. This first step of flattening the data is needed in order to add the filters. 
- closes #219

## Testing Instructions
- visit the cip planner: https://ccfp-asset-d-feature-fl-3knsmw.herokuapp.com/
  - login with username=`test` / password=`supersecret`
- build a portfolio. see how the columns are different and the phase funding stream make each row unique
- export the portfolio and make sure the data looks right


xmedr (commented):
Things look good for the table! I have a couple questions in there but also wanted to flag some bugs. Not sure if we want to address these in this pr, or approve this and open a different one.

#### Creating Portfolio
I see the save bug you mentioned. When I was first messing around with the portfolio, I added and removed things before I gave it a name and tried to save it. A cancel button then appears to the right of the save button (shown below), and when I navigate away to a different portfolio and try to look for it again, it isn't present in the select dropdown. If you enter the name and save *before* you select any fundings, the portfolio gets created.

image
---

#### Saving Portfolio
However, actually saving the portfolio after adding fundings is also hit or miss. Below is what it looks like when i've made the portfolio, added fundings, and tried to hit save. It thinks I want to edit the title again and test 2 gets unselected from the select dropdown. Hitting save doesn't do anything. Cancel removes the name editing and displays the usual "All changes saved" message, but after refreshing, it doesn't look like it has saved.

image
---

#### Exporting Data
For the first portfolio I created, I got a weird export before saving it. It had the items I had added, but also some extras that I couldn't find in either table. This I haven't been able to reproduce, so maybe it was a fluke.

asset_dashboard/static/js/components/table_utils/projectColumns.js
```diff
@@ -22,69 +22,75 @@ const projectColumns = (selector) => {
       accessor: 'name'
     },
     {
-      Header: 'Description',
-      accessor: 'description', 
+      Header: 'Estimated Cost',
+      accessor: 'budget',
+      Cell: props => props.value.toLocaleString() // adds commas
```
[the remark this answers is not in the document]
smcalilly: Yeah that's a feature of the [react-table package](https://www.npmjs.com/package/react-table).

It looks like the maintainer of that library did a major rebrand and now the docs are very confusing, but I think this might be it? https://tanstack.com/table/v8/docs/api/core/cell

Otherwise there are a ton of stackoverflows that can show you how you can use that, like: https://stackoverflow.com/questions/45968791/react-table-custom-cell-component-that-references-several-row-properties

asset_dashboard/migrations/0045_auto_20230817_2121.py
smcalilly, line 1: @xmedr this is the custom migration i had to write. i added the model field in the above migration and then did an "empty" migration so i could write python code to update the existing data.

smcalilly:
@xmedr i fixed that bug. can you please review again?

asset_dashboard/migrations/0045_auto_20230817_2121.py
xmedr: Ah so this will run sequentially with the other migrations and update, this is neat!

xmedr (commented):
Hm, the interface works now, but the saving is still being a bit strange. I'm testing using my previous "test 2" portfolio.

I can select a few and hit save, and we get the expected ui outputs, but upon refreshing the page we get different fundings than what I had selected. The export shows the new, incorrectly saved fundings.

Before saving
image

After refreshing
image

I also tried making a new portfolio "test 3", adding phases, and saving the name afterwards like before. That works now and the portfolio is created, but it's got the same saving weirdness that I described above

smcalilly:
@xmedr thanks for your QA. i fixed that. lemme know if you find any other issues.

asset_dashboard/serializers.py
```diff
@@ -61,6 +61,9 @@ def update(self, instance, validated_data):
 
         phase_data = validated_data.get('phases', [])
 
+        print('phase_data', phase_data)
+        print('instance', instance)
+
```
xmedr, line 66: did we want to keep these prints in here for logging in production or was this just for debugging?

xmedr (approved):
nice, that did it!! everything is working as expected now 🎉

just got one question about print statements and you can bring this in

asset_dashboard/serializers.py
smcalilly: oop good catch, thanks.

Issue #288, opened 2025-03-31
Add additional filters to CIP Planner page


jim-z:
Add filter for phase

![Image](https://github.com/user-attachments/assets/2eb70edf-60a4-4d9c-a10e-5e64ab33ca1f)

Perhaps UI could be rearranged as "Filter 1", "Filter 2",  "Filter 3" with a dropdown to select table field and then a conditional dropdown next to it to select specific field value.

xmedr:
@jim-z Having a phase filter makes sense! Though, could you say more about the second part of this issue? 

It sounds like the goal is to let you folks decide what filters you'd like to see on the page in order to make the filters themselves more flexible, and to have more filtering options, is that right? If so, I could see a list of toggle-able filters being useful.

jim-z:
@xmedr yeah, that's the goal...having the ability to filter on any available field in the underlying database.  Might be a heavy lift/too significant a UI change but example of how filtering could potentially work here: 
https://datacatalog.cookcountyil.gov/Property-Taxation/Assessor-Property-Tax-Exempt-Parcels/vgzx-68gb/explore/query/SELECT%0A%20%20%60pin%60%2C%0A%20%20%60year%60%2C%0A%20%20%60township_name%60%2C%0A%20%20%60township_code%60%2C%0A%20%20%60owner_num%60%2C%0A%20%20%60owner_name%60%2C%0A%20%20%60class%60%2C%0A%20%20%60property_address%60%2C%0A%20%20%60property_city%60%2C%0A%20%20%60lon%60%2C%0A%20%20%60lat%60/page/filter

xmedr:
@jim-z Ah sweet, thanks for the example! We'll get started on this and keep you updated, but for now we have a change to include the phase filter coming.

PR #290, opened 2025-04-18
Add phase filter, display readable labels for funding sources

xmedr:
## Overview

This branch adds a phase filter to the CIP planner. It also changes the values in the table for funding sources to be a more readable label.

Because of that change, we're now ensuring that funding source (and phase) filtering happens based on the comparison between **label** in the table (i.e. Grants, Fees, & Other) and the **label** of the dropdown option (i.e. Grants, Fees, & Other), instead of the **label** in the table (i.e. Grants, Fees, & Other) and the **value** of the dropdown option (i.e. grants_fees_other).

- Connects #288 

### Demo

image

### Notes

This does not close the connected issue because there's going to be an extra step to select which filters users want to use dynamically. But I wanted to at first get a pattern in for phase filtering.

## Testing Instructions
* Head to the CIP planner
* Mess with the filters and confirm they work as expected
  * Pay special mind to the new phase filter and the funding source filter

### DIFF
--- a/asset_dashboard/static/js/PortfolioPlanner.js
+++ b/asset_dashboard/static/js/PortfolioPlanner.js
@@ -36,7 +36,8 @@ class PortfolioPlanner extends React.Component {
       selectedSection: '',
       selectedFundingSecured: '',
       selectedFundingSource: '',
-      selectedYear: ''
+      selectedYear: '',
+      selectedPhase: ''
     }
 
     this.searchProjects = this.searchProjects.bind(this)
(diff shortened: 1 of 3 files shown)


asset_dashboard/static/js/PortfolioPlanner.js
```diff
@@ -81,7 +88,7 @@ class PortfolioPlanner extends React.Component {
           senate_districts: project.senate_districts,
           commissioner_districts: project.commissioner_districts,
           key: funding.id,
-          funding_source: funding['source_type'],
+          funding_source: fundingSourceLabel,
```
xmedr, line 91: Making sure to show the readable label instead of the snake_cased one on the table

asset_dashboard/static/js/PortfolioPlanner.js
```diff
@@ -513,14 +534,26 @@ class PortfolioPlanner extends React.Component {
     }
   }
 
+  filterPhase(project) {
+    if (this.state.selectedPhase) {
+      // Ensure we're matching on the phase value displayed on the table
+      const selectedOption = this.props.phaseOptions.filter(
+        (source) => source['value'] == this.state.selectedPhase
+      )[0]
+      return project.phase == selectedOption.label
```
xmedr, line 543: Decided to not use the `within()` helper function here (and by extension in for the funding sources) so that we could get an exact match and not have `Construction` match with both `Construction` and `Construction Engineering`.

antidipyramid (approved):
Works great!

smcalilly:
I'm still subscribed to this repo lol. Miss y'all!!! Hope this codebase doesn't make you struggle too much :1st_place_medal:

xmedr:
@smcalilly loll hi, miss you too and I hope you're doing well!! I've relayed to folks the horrors that you and I experienced with formsets, so know that we didn't suffer in vain there 😤

Issue #293, opened 2025-04-21
Update phase and status domain options


jim-z:
![Image](https://github.com/user-attachments/assets/52bdef88-41a9-4e0e-93e7-d8b2bad2b5d2)

Update status options to include: inactive/hold pending funding, complete, active/in progress.  Need to confirm but I don't think there is any need for "new".  Existing "ongoing" entries = "active/in progress".

![Image](https://github.com/user-attachments/assets/94133f73-62c7-49e2-85c5-f2be4ad6d3a6)

Under phase options, add a category for "conditions assessment/inspection"

PR #294, opened 2025-05-13
Update phase filters + add new phase type

antidipyramid:
## Overview

Connects #293 

### Demo
Screenshot 2025-05-13 at 9 56 05 AM
Screenshot 2025-05-13 at 9 56 20 AM

### Notes

I added a data migration to move phases that had previous statuses of "ongoing" to "active/in progress".

## Testing Instructions

* Verify that the new phase filter/type is available


asset_dashboard/migrations/0047_update_ongoing_phases_to_active.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 3.2.19 on 2025-05-13 13:34
+
+from django.db import migrations
+
+
+def forwards_func(apps, schema_editor):
+    Phase = apps.get_model("asset_dashboard", "Phase")
+    for phase in Phase.objects.all():
```
antidipyramid, line 8: Change status of any `ongoing` statuses to `active`.

antidipyramid:
@jim-z We can change any phases with a status of `ongoing` to `active`. Do we want to do something similar for phases with a `new` status?

jim-z:
@antidipyramid yes, I think it makes sense to treat existing "ongoing" and "new" the same and convert both to "active"

xmedr (approved):
can confirm that the new status and type options work, as did the data migration!

would you be able to make one more quick change to the phase table to show the display values for the new status options? it'd likely be a small change to this line
https://github.com/fpdcc/ccfp-asset-dashboard/blob/c7629f4299b4e35223b4e0ea0c6b3f5eb926a50b/asset_dashboard/templates/asset_dashboard/partials/phase_table.html#L21

to instead be
```html
<td>{{ phase.get_status_display }}</td>
```

PR #299, opened 2025-05-20
Rearrange csv fields, add recently created fields

### LINKED ISSUE #297: updates to csv export function
1.) rearrange default field export order so that project_ID, phase_id, phase_funding_id, name appear at the beginning of csv table

2.) add missing fields to csv export: 

Actual Cost

Image

Requester and Project Status

Image

xmedr:
## Overview

The `project_ID`, `phase_id`, `phase_funding_id`, and `name` fields have been moved to the start of the csv, and a phase's `actual_cost`, `project_status`, and `project_requester` have been added.

- Closes #297

## Testing Instructions

* Pull down this branch and log in
  * Make sure you have one or more projects with phases, a status, and a requester
* Head to the CIP Planner and add projects to a portfolio
* Export the portfolio as a csv
* Confirm that the adjustments described above are present
* Confirm that all affected columns are populated as expected

### DIFF
--- a/asset_dashboard/static/js/PortfolioPlanner.js
+++ b/asset_dashboard/static/js/PortfolioPlanner.js
@@ -94,12 +94,15 @@ class PortfolioPlanner extends React.Component {
           funding_secured: funding['funding_secured'] ? 'Yes' : 'No',
           phase_year: project.phase_year,
           estimated_bid_quarter: project.estimated_bid_quarter,
-          status: project.status,
+          phase_status: project.phase_status,
+          project_status: project.project_status,
           project_manager: project.project_manager,
           countywide: project.countywide,
           assets: project.assets,
           project_id: project.project_id,
           phase_id: project.pk,
+          actual_cost: funding.actual_cost,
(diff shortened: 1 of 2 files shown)


asset_dashboard/static/js/PortfolioPlanner.js
```diff
@@ -589,7 +595,10 @@ class PortfolioPlanner extends React.Component {
         'category': project.category,
         'project_manager': project.project_manager,
         'phase': project.phase,
-        'status': project.status,
+        'phase_status': project.phase_status,
+        'actual_cost': project.actual_cost,
+        'project_status': project.project_status,
```
xmedr, line 600: Just an fyi that the terminology used in the planner is a little confusing.

It looks like each row describes a separate project or maybe every project's phase, but each row is actually every funding stream, connected to a phase, that belongs to a project. So there's a little bit of a disconnect between the variable names in the view vs. what's in here.

I renamed this field to help with clarity just a bit. I'm sure there'll be other spots we come back to for clarity's sake.

PR #304, opened 2025-07-14
Add mgmt cmd to redo project geographies

### LINKED ISSUE #301: Try re-calculating project geographies

### LINKED ISSUE #302: Management command to re-populate geographic info
Related to #301

xmedr:
## Overview

Small pr to add a command that saves select local assets, meant to run on a regular basis so all projects' geographies can stay up to date. Only one asset per project is saved in order to cut down on processing time.

This also includes an addition to the README detailing how to shell in to an EC2 instance.

- Closes #301 
- Connects #302

### Notes

I tested saving every single asset on the staging site, and that took about 20 minutes for ~2200 assets. This method of just saving one asset per project takes about 2 minutes and still updates every phase/project with an asset, so I went with that. I don't believe this would update the assets themselves on a regular basis from the remote db, but is a start.

I'd also need to do some research on how to set up a cronjob on AWS if/when this comes in. If you have any experience with that, that would be huge!

## Testing Instructions

* Edit a project's phase to give it an asset, or find a project with assets and no geographies.
  * If all projects with assets have geographies, shell in and [`clear()`](https://docs.djangoproject.com/en/5.2/ref/models/relations/#django.db.models.fields.related.RelatedManager.clear) all zones from one or two projects.
* Run `docker compose run --rm app python manage.py update_project_geos`
  * Confirm that it runs in a timely manner and populates project geographies
  * If no geographies get populated, run `docker compose run --rm app make districts` before trying the update command again.


requirements.txt
```diff
@@ -10,6 +10,7 @@ django-datatables-view==1.19.1
 django-widget-tweaks==1.4.8
 djangorestframework==3.13.1
 djangorestframework-gis==0.18
+tqdm==4.67.1
```
xmedr, line 13: Lil progress bar

antidipyramid (approved):

PR #303, opened 2025-06-26
Autosave and warn on unsaved phase/funding edits

xmedr:
## Overview

This branch autosaves phase and funding stream details whenever a change happens in one of the inputs. It also extends the warning that was already in place for unsaved funding streams, to also apply to phase details.

- Connects #276 

### Demo

image

## Testing Instructions

* Pull down this branch, log in, and head to a created project
  * If needed, add a new phase and/or funding streams to a phase
* Edit the phase details and funding stream details
  * Confirm any changes to either section causes the page to save changes on both
  * Confirm that they get saved by refreshing the page, or viewing the detail of the phase from the project view
  * Confirm that their save statuses, indicated by the icon to the right of the header for each card, get updated independently
* Trigger an error in one of the forms. You can, for example, insert an `e` into one of the year inputs
  * Confirm that the error icon appears just for the section that has an error
  * Try navigating away without fixing the error and confirm that a warning displays
* Click the "set phase as countywide" checkbox towards the bottom
  * Confirm that a success message shows and the page reloads to show the grayed out phase assets section as expected.
* Try to add a new phase
  * Confirm that none of the icons, auto saving, or warnings happen when creating a new phase from scratch

PR #307, opened 2025-10-20
Fix app making multiple fundings for each new one made

### LINKED ISSUE #306: Phase funding amount bug
After adding a phase and entering a value for funding, app is creating multiple lines in error:

Image

View of overview page after updating funding:

Image

Results after clicking edit:

Image

Should only be displaying the line circled red

xmedr:
## Overview

This branch stops the app from creating multiple identical funding streams any time a **new** one is made. It does this by giving the added form the pk of the newly made funding stream after creation. In my first iteration of this (#303) I only tested existing funding streams and missed this intricacy.

- Closes #306

### Notes

I also added django debug toolbar into the app in this branch that'll only run when debug is `True` so that I could investigate why refreshing the phase edit page took so long, but the sql query times that the toolbar reported seemed pretty short. I can look more into that more another time.

## Testing Instructions

* Build the container so that the debug toolbar gets installed into your instance, then bring it up
* Head to a project phase, or create one if necessary
* Add one or more new phase funding stream(s)
  * Confirm that the save icon at the top of the funding section accurately shows that it has been saved
  * Confirm that the "Remove" button on the funding has been relabeled as "Delete" after saved, and works as expected
  * Refresh the phase edit page and confirm that there is only one copy of the funding stream you've made
  * Confirm that editing an existing funding stream still works as expected
* Confirm that the django debug toolbar works as expected

### DIFF
--- a/asset_dashboard/settings.py
+++ b/asset_dashboard/settings.py
@@ -44,6 +44,20 @@ if os.getenv('SENTRY_DSN'):
         integrations=[DjangoIntegration()],
     )
 
+if DEBUG:
+    import socket
+
+    # Add dynamically generated Docker IP
+    # Don't do this in production!
+    hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
+    INTERNAL_IPS = [ip[:-1] + "1" for ip in ips] + ["127.0.0.1"]
+
+# Django Debug Toolbar Panel Settings
+DEBUG_TOOLBAR_PANELS = [
+    "debug_toolbar.panels.sql.SQLPanel",
+    "debug_toolbar.panels.cache.CachePanel",
+]
+
 # Application definition
 
 INSTALLED_APPS = [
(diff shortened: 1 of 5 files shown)


asset_dashboard/settings.py
```diff
@@ -44,6 +44,20 @@
...
+    # Don't do this in production!
+    hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
+    INTERNAL_IPS = [ip[:-1] + "1" for ip in ips] + ["127.0.0.1"]
+
+# Django Debug Toolbar Panel Settings
+DEBUG_TOOLBAR_PANELS = [
+    "debug_toolbar.panels.sql.SQLPanel",
+    "debug_toolbar.panels.cache.CachePanel",
+]
+
```
xmedr, line 60: Lifted from metro

asset_dashboard/templates/asset_dashboard/partials/forms/add_edit_phase_form.html
```diff
@@ -379,6 +383,17 @@ <h3>Phase Assets</h3>
             }).then(data => {
               if (!backendValid) {
                 showBackendErrors(data, ...errors)
+              } else if (!funding.id) {
+                // Associate this new form with its new funding stream
+                funding.id = data["id"]
```
xmedr, line 388: The crux of the fix. Once the serializer is able to give us the newly created funding's id, we add that to the form so that any subsequent submissions of this funding have the id attached and don't create a new one.

antidipyramid (approved):
Looks great!

xmedr (commented):
Thanks!

Posting a couple comments that I thought I'd already posted

